PR #5, opened 2023-11-24
add login screen and store, type, i18n boilerplate

bkmaibach:


bkmaibach:
Opened integration/auth PR instead

PR #7, opened 2023-11-27
Integration branch for auth related features

bkmaibach:


app/views/devise/confirmations/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<h2>Resend confirmation instructions</h2>
```
LaurelOlson, line 1: We shouldn't need any of the devise views - can just use the defaults as they will be overridden eventually by some sort of template mailer most likely. Did you need to customize something?

db/seeds.rb
```diff
@@ -7,3 +7,7 @@
 #   ["Action", "Comedy", "Drama", "Horror"].each do |genre_name|
 #     MovieGenre.find_or_create_by!(name: genre_name)
 #   end
+
+User.create(email: "user@laterolabs.com", password: "password")
+
+User.create(email: "admin@example.com", password: "password")
```
LaurelOlson, line 13: These should use `first_or_create` and should also set `username` to avoid uniq index errors

db/schema.rb
```diff
@@ -26,16 +26,30 @@
     t.index ["user_id"], name: "index_allowlisted_jwts_on_user_id"
   end
 
+  create_table "test_tables", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "name"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
```
LaurelOlson, line 34: cruft - if you drop your DB and remigrate this should go away

app/frontend/styles/theme/components/button.ts
```diff
@@ -61,5 +61,14 @@ export const Button = {
...
     },
+    link: {
+      color: "text.link",
+      fontWeight: "normal",
+      _hover: {
+        textDecoration: "underline",
+        _disabled: { textDecoration: "none" },
+      },
+      _disabled: { ...disabledStyles, bg: "inherit" },
+    },
```
LaurelOlson, line 72: I think this is the same as the `tertiary` button

app/frontend/stores/user-store.ts
```diff
@@ -0,0 +1,49 @@
+import { values } from "mobx"
+import { Instance, flow, types } from "mobx-state-tree"
+import * as R from "ramda"
+import { withEnvironment } from "../lib/with-environment"
+import { withRootStore } from "../lib/with-root-store"
+import { IUser, UserModel } from "../models/user"
+// import { IUser, UserModel } from "../models/user"
```
LaurelOlson, line 7: cruft

app/frontend/stores/user-store.ts
```diff
@@ -0,0 +1,49 @@
...
+      return values(self.usersMap) as IUser[]
+    },
+  }))
+  .views((self) => ({
+    getUser(id) {
+      return self.users.find((u) => u.id == id)
+    },
+    isCurrentUser(user) {
+      return user.id == self.currentUser.id
+    },
```
LaurelOlson, line 29: are these being used? Let's avoid adding extra code unless it's needed as this increases the chance of duplications / stale code. Same with the actions below.

app/frontend/stores/session-store.ts
```diff
@@ -0,0 +1,69 @@
...
+      const response: any = yield self.environment.api.validateToken() // now try to validate this with the server
+      self.handleLogin(response)
+    }),
+    validateCredentials: flow(function* (email, password) {
+      const response: any = yield self.environment.api.login(email, password)
+      if (response.ok) {
+        if (!self.loggedIn) yield self.environment.api.logout()
+        return response.data.data
+      }
+    }),
```
LaurelOlson, line 46: remove this - not used

app/frontend/stores/session-store.ts
```diff
@@ -0,0 +1,69 @@
...
+    },
+    handleForgotPasswordRequest: flow(function* (params) {
+      const response = yield self.environment.api.requestPasswordReset(params)
+      if (response.ok) {
+        return true
+      }
+    }),
+  }))
+  .actions((self) => ({
+    validateToken: flow(function* () {
```
LaurelOlson, line 36: we should call `validateToken` in a `useEffect` on the navigation component to check and see if there is an authenticated user (e.g. once logged in, if the page is reloaded the user should not have to re-login OR if the cookie is invalid, the user should be logged out and see the login page) - see PM for reference

app/frontend/components/shared/ui/shared-spinner.tsx
```diff
@@ -0,0 +1,8 @@
+import { Spinner, SpinnerProps } from "@chakra-ui/react"
+import React from "react"
+
+interface ISharedSpinner extends SpinnerProps {}
+
+export const SharedSpinner = ({ ...rest }: ISharedSpinner) => {
+  return <Spinner {...rest} />
+}
```
LaurelOlson, line 8: this seems quite unnecessary as it's just wrapping the chakra Spinner?

LaurelOlson (changes requested):

app/controllers/api/passwords_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  # POST /resource/password
+  def create
+    self.resource = User.find_by_username(resource_params[:username])
+    if self.resource
+      self.resource.send_reset_password_instructions
+      if successfully_sent?(resource)
+        render json: {}, status: :ok and return
+      else
+        # TODO: messaging
+        render json: {}, status: 400 and return
```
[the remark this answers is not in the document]
bkmaibach: this will be done once messaging is integrated

app/controllers/api/confirmations_controller.rb
```diff
@@ -0,0 +1,37 @@
+class Api::ConfirmationsController < Devise::ConfirmationsController
```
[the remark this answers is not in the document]
bkmaibach: This will be handled once this is rebased on flash message code

app/controllers/api/passwords_controller.rb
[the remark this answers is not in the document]
bkmaibach: This will be handled once this is rebased on flash message code

app/views/devise/confirmations/new.html.erb
bkmaibach: see:
app/views/devise/mailer/reset_password_instructions.html.erb

app/frontend/styles/theme/components/button.ts
bkmaibach: The color is different. Should I:
- Manually specify the color on the component?
- change the tertiary component?
- ignore the color?

app/frontend/components/shared/ui/shared-spinner.tsx
bkmaibach: This is a stub for when we want a site wide standardized spinner later, let's leave it for now?

app/frontend/components/shared/form/email-form-control.tsx
```diff
@@ -0,0 +1,35 @@
+import { Flex, FormControl, FormControlProps, FormErrorMessage, FormLabel, Input, InputGroup } from "@chakra-ui/react"
+import React from "react"
+import { useFormContext } from "react-hook-form"
+import { useTranslation } from "react-i18next"
+
+interface IEmailFormControlProps extends FormControlProps {
+  validate?: boolean
```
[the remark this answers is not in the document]
bkmaibach: I dont think it makes sense to validate the password when logging in, I don't usually see that. Same with username,

app/frontend/stores/session-store.ts
bkmaibach: Done and done

app/frontend/components/shared/form/email-form-control.tsx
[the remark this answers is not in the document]
LaurelOlson: OK looks like with the regex fix you are now always validating so will need to add that check back in so it doesn't validate on the login form

app/views/devise/confirmations/new.html.erb
LaurelOlson: OK let's remove the other views then and add a TODO comment in the reset password mailer html to remove once we add in templates

app/frontend/components/shared/center-container.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const CenterContainer = ({ children }: ICenterContainerProps) => {
+  return (
+    <Container maxW="container.md" h="full" p={0}>
+      <Flex direction="column" height={`calc(100vh - ${NAVBAR_HEIGHT})`} justify={{ base: "flex-start", md: "center" }}>
+        {children}
+      </Flex>
+    </Container>
+  )
+}
```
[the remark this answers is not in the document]
LaurelOlson: Did you end up still needing this after our discussion yesterday?

app/frontend/components/shared/form/email-form-control.tsx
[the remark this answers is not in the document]
bkmaibach: @LaurelOlson I cant seem to see the behaviour you are describing?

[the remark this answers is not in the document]
LaurelOlson: inside matchesEmailRegex - you don't use the `validate` prop at all in this component. It should use the `!validate` condition similar to the password / username inputs

PR #14, opened 2023-11-30
Keithmaibach/sc 66723/create landing page

bkmaibach:


package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@bcgov/bc-sans": "^2.1.0",
+    "@chakra-ui/icons": "^2.1.1",
```
LaurelOlson, line 29: Karen mentioned the gov uses icons from Font Awesome 6  Regular so we should add that in and import icons from there

LaurelOlson (changes requested):

PR #33, opened 2024-01-03
add models for permit classifications

bkmaibach:
## Description

Adds models, factories, seeders for the permit classification STI which will let us create an arbitray set of permit classifications. Currently there is PermitType and Activity (aka WorkType)

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://app.shortcut.com/codechallengell/story/66489/setup-building-permit-options

## Steps to QA

Set up the db and seed with rails db:seed

This should create a set of PermitTypes and Activities, and some RequirementTemplates to use them.

No frontend in this PR.

## General Checklist

- [x ] 📝 Use descriptive commit messages
- [ x] 📗 Update any related documentation and include any relevant screenshots
- [x ] 📱 For UI-related tasks, ensure responsive design is implemented across multiple screen size
- [ x] 🗂️ Move your relevant story/task to the _Ready for Code Review_ state

### DIFF
--- a/app/models/requirement_template.rb
+++ b/app/models/requirement_template.rb
@@ -1,0 +1,14 @@
+class RequirementTemplate < ApplicationRecord
+  belongs_to :activity, class_name: "Activity"
+  belongs_to :permit_type, class_name: "PermitType"
+
+  validates :activity_id, uniqueness: { scope: :permit_type_id }
+  validate :ensure_correct_type_for_permit_type_and_activity
+
+  private
+
+  def ensure_correct_type_for_permit_type_and_activity
+    errors.add(:permit_type, "must be of type PermitType") unless permit_type.type == "PermitType"
+    errors.add(:activity, "must be of type Activity") unless activity.type == "Activity"
+  end
+end
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -1,0 +1,2 @@
+class Activity < PermitClassification
+end
--- a/app/models/permit_application.rb
+++ b/app/models/permit_application.rb
@@ -5,13 +5,6 @@ class PermitApplication < ApplicationRecord
   # Custom validation
   validate :submitter_must_have_role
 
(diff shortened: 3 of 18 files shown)


app/models/requirement_template.rb
```diff
@@ -0,0 +1,14 @@
+class RequirementTemplate < ApplicationRecord
+  belongs_to :activity, class_name: "Activity"
+  belongs_to :permit_type, class_name: "PermitType"
```
taw27, line 3: does the classname need to be specified? Asking because we have something similar in productmotion where TreeItem is the base class and Feature and Collection inherit from it. It seems like rails is able to automatically handle this in pm code base, so wondering if the same can be achieved here

app/models/requirement_template.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  validates :activity_id, uniqueness: { scope: :permit_type_id }
+  validate :ensure_correct_type_for_permit_type_and_activity
+
+  private
+
+  def ensure_correct_type_for_permit_type_and_activity
+    errors.add(:permit_type, "must be of type PermitType") unless permit_type.type == "PermitType"
+    errors.add(:activity, "must be of type Activity") unless activity.type == "Activity"
+  end
```
taw27, line 13: @bkmaibach does it work without these validations? Asking because we have something similar in productmotion where TreeItem is the base class and Feature and Collection inherit from it. It seems like rails is able to automatically handle this in pm code base, so wondering if the same can be achieved here

taw27 (approved):

app/models/requirement_template.rb
bkmaibach: probably

app/models/requirement_template.rb
bkmaibach: probably

PR #41, opened 2024-01-09
Feature/create seeded template/sc 67048

sunto:
## Description

The purpose of this is to allow seeding of requirements straight from the excel file that the HOUS team is developing.
All devs must ensure they have git lfs set up on their system for this to work, as this permanently puts that seed file through gitlfs.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [x] 📦 Chore (Release)
- [ ] ✅ Test

Please use this format to link: Implements Story [67048](https://app.shortcut.com/codechallengell/story/67048/create-seeded-template).

## Steps to QA

Should be able to run the following seed command now:
```
PermitClassificationSeeder.seed
RequirementsFromXlsxSeeder.seed

RequirementTemplate.first.to_form_json
```
This should add requirements to the backend

## UI Accessibility Checklist
does not apply, backend

## General Checklist

- [x] ✅ Provide tests for your changes where relevant
- [x] 📗 Update any related documentation and include any relevant screenshots
- [x] 📱 For UI-related tasks, ensure responsive design is implemented across multiple screen size

### DIFF
--- a/app/models/requirement.rb
+++ b/app/models/requirement.rb
@@ -2,12 +2,22 @@ class Requirement < ApplicationRecord
   has_many :requirement_block_requirements, dependent: :destroy
   has_many :requirement_blocks, through: :requirement_block_requirements
 
-  enum input_type: { textfield: 0, number: 1, checkbox: 2, select: 3, multi_option_select: 4, date: 5 }, _prefix: true
+  enum input_type: {
+         text: 0,
+         number: 1,
+         checkbox: 2,
+         select: 3,
+         multi_option_select: 4,
+         date: 5,
+         textarea: 6,
+       },
+       _prefix: true
 
+  before_create :set_requirement_code
   validate :validate_options_for_select_inputs
 
   DEFAULT_FORMIO_TYPE_TO_OPTIONS = {
-    textfield: {
+    text: {
       type: "simpletextfield",
     },
     number: {
@@ -85,6 +95,11 @@ class Requirement < ApplicationRecord
   end
 
   private
+
+  # requirement codes should not be auto generated during seeding.  Use uuid if not provided
+  def set_requirement_code
+    self.requirement_code ||= SecureRandom.uuid
+  end
 
   def formio_type_options
     DEFAULT_FORMIO_TYPE_TO_OPTIONS[input_type.to_sym] || {}
--- a/app/models/requirement.rb
+++ b/app/models/requirement.rb
@@ -2,14 +2,12 @@ class Requirement < ApplicationRecord
   has_many :requirement_block_requirements, dependent: :destroy
   has_many :requirement_blocks, through: :requirement_block_requirements
 
-  enum input_type: { text: 0, number: 1, checkbox: 2, select: 3, multi_option_select: 4, date: 5 }, _prefix: true
+  enum input_type: { textfield: 0, number: 1, checkbox: 2, select: 3, multi_option_select: 4, date: 5 }, _prefix: true
 
   validate :validate_options_for_select_inputs
 
-  before_validation :set_requirement_code
-
   DEFAULT_FORMIO_TYPE_TO_OPTIONS = {
-    text: {
+    textfield: {
       type: "simpletextfield",
     },
     number: {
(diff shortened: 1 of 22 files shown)


sunto:
Please DO NOT merge until sections are added.

sunto:
Sections now added

app/models/requirement.rb
```diff
@@ -2,14 +2,12 @@ class Requirement < ApplicationRecord
   has_many :requirement_block_requirements, dependent: :destroy
   has_many :requirement_blocks, through: :requirement_block_requirements
 
-  enum input_type: { text: 0, number: 1, checkbox: 2, select: 3, multi_option_select: 4, date: 5 }, _prefix: true
+  enum input_type: { textfield: 0, number: 1, checkbox: 2, select: 3, multi_option_select: 4, date: 5 }, _prefix: true
```
taw27, line 5: any reason to change it to `textfield`? Seems a bit of a mismatch with the other types as they do not have field in their names.

sunto:
To test, must run seeding, verify a requirement template with sections is created.
```
PermitClassificationSeeder.seed
RequirementsFromXlsxSeeder.seed

RequirementTemplate.first.to_form_json
```

app/models/requirement_block.rb
```diff
@@ -4,6 +4,9 @@ class RequirementBlock < ApplicationRecord
   has_many :requirement_block_requirements, -> { order(position: :asc) }, dependent: :destroy
   has_many :requirements, through: :requirement_block_requirements
 
+  has_many :requirement_template_section_requirement_blocks, dependent: :destroy
+  has_many :requirement_templates, through: :requirement_template_requirement_blocks
```
taw27, line 8: Cannot fin this relationship `:requirement_template_requirement_blocks`

app/services/requirements_from_xlsx_seeder.rb
```diff
@@ -0,0 +1,132 @@
+class RequirementsFromXlsxSeeder
+  def self.seed
+    file_name = "#{Rails.root}/db/templates/Core Permit Requirements List.xlsx"
```
taw27, line 3: can be changed later, but `file_name` can probably be accepted as a param to the static method with the default param as the value

app/models/requirement.rb
```diff
@@ -92,11 +90,6 @@ def formio_type_options
     DEFAULT_FORMIO_TYPE_TO_OPTIONS[input_type.to_sym] || {}
   end
 
-  # TODO: requirement codes should be auto generated. Using a uuid for now. Later we have to replace it with a more human readable code
-  def set_requirement_code
-    self.requirement_code = SecureRandom.uuid
-  end
-
```
taw27, line 99: This should probably be still included, as other requirements will need it, which are not seeded.

def set_requirement_code
    self.requirement_code ||= SecureRandom.uuid
  end

app/services/requirements_from_xlsx_seeder.rb
```diff
@@ -0,0 +1,132 @@
...
+      .drop(3)
+      .compact
+      .uniq
+      .each do |section_name|
+        rs =
+          requirement_template
+            .requirement_template_sections
+            .where(name: section_name)
+            .first_or_create!(name: section_name)
+        rs.update(position: rs_position_incrementer)
```
taw27, line 86: is this to place section at top in reverse order?

db/schema.rb
```diff
@@ -60,6 +60,8 @@
   end
 
   create_table "permit_applications", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.integer "permit_type", default: 0
+    t.integer "building_type", default: 0
```
taw27, line 64: Might need to drop database and remigrate. I think keith removed these fields, so it might be getting picked up from existing db fields

app/models/requirement_block.rb
sunto: typo. will fix.  good catch.

app/services/requirements_from_xlsx_seeder.rb
sunto: its placed in the order they are in from the excel file.  But it seeds the section positioning.

app/models/requirement.rb
taw27: Yup sure. I added to before validation in case we add validations related to the code

taw27 (approved):

PR #70, opened 2024-01-26
add optimized user archive, restore, and user sign in tracking

bkmaibach:
## Description

Adds user archive ad restore
Refactors user search
Add User last sign in tracking

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-232?atlOrigin=eyJpIjoiMDllYWRhNGNiMDAyNDZjZmE5NDNhOGI5MmFlZDY2NWQiLCJwIjoiaiJ9

## Steps to QA

Log in as super admin, log in to jurisdiction, archive users from manage menu
These users will appear in the archived list (see by clicking button at the bottom)

Notice that there are no repeated search or fetchJurisdiction calls! 🔥 

## UI Accessibility Checklist

No new major UI changes

## General Checklist

- [x ] 📝 Use descriptive commit messages
- [x ] 📗 Update any related documentation and include any relevant screenshots
- [x ] 📱 For UI-related tasks, ensure responsive design is implemented across multiple screen size
- [x ] 🗂️ Move your relevant story/task to the _Ready for Code Review_ state


app/controllers/api/invitations_controller.rb
```diff
@@ -19,12 +19,14 @@ def remove
 
   def update
     raw_invitation_token = update_resource_params["invitation_token"]
-    self.resource = accept_resource
+    user = User.find_by_invitation_token(raw_invitation_token, true)
```
sunto, line 22: This seems a little weird. What sort of errors are we getting?

app/frontend/stores/session-store.ts
```diff
@@ -6,7 +6,7 @@ export const SessionStoreModel = types
   .model("SessionStoreModel")
   .props({
     loggedIn: types.optional(types.boolean, false),
-    isValidating: types.optional(types.boolean, false),
+    isValidating: types.optional(types.boolean, true),
```
sunto, line 9: how come this is true by default?

app/frontend/stores/user-store.ts
```diff
@@ -31,7 +35,7 @@ export const UserStoreModel = types
   }))
   .actions((self) => ({
     setUsers(users) {
-      R.forEach((u) => self.usersMap.put(u), users)
+      R.map((u) => self.usersMap.put(u), users)
```
taw27, line 38: It's better to use forEach over map in this case. Map creates unnecessary memory which is not being used anywhere

app/frontend/models/jurisdiction.ts
```diff
@@ -1,76 +1,43 @@
...
+    reverseQualifiedName: types.string,
+    localityType: types.string,
+    qualifier: types.string,
+    reviewManagersSize: types.number,
+    reviewersSize: types.number,
+    permitApplicationsSize: types.number,
+    description: types.maybeNull(types.string),
+    // JSONB data type can be represented as a frozen type
+    checklistSlateData: types.maybeNull(types.frozen<Descendant[]>()),
+    lookOutSlateData: types.maybeNull(types.frozen<Descendant[]>()),
```
taw27, line 24: We should refactor this to use quilljs which we can just store the raw html. If not handled in this PR then please make a separate card for it. We should eventually remove slate from the app

app/frontend/stores/session-store.ts
taw27: @sunto this true by default to prevent race conditions.

app/controllers/api/invitations_controller.rb
bkmaibach: What happens here is that because the user is accepted and obtained at the same time, if the user does not already have a first and last name it will fail validation. The user must have its first and last name set before confirmation.

app/frontend/stores/session-store.ts
bkmaibach: @sunto can confirm, tawshif and I worked this out.
basically we dont want to show the logged in pages until we are actually sure the user is logged in, which depends on the isValidating flags. Leaving this as default false means that the useEffect hooks in logged in versions of the same route will run and then again on the logged in page.

app/frontend/stores/user-store.ts
bkmaibach: roger

app/frontend/models/jurisdiction.ts
bkmaibach: I will do this as a part of the edit jurisdictiion card

taw27 (approved):

PR #77, opened 2024-02-01
Feature/bphh 52/add save and compliance flow

sunto:
## Description
allows persistence of back end form to front end

- [ x] 🍕 Feature

Precursor for doing automatec code compliance testing

## Steps to QA

Log in as a submitter, go to a seeded permit application, update a few answers, hit 'submit'
Refresh the screen on permit applications index, go back to the same permit applicaiton, the answers should load.

### DIFF
--- a/app/frontend/components/shared/chefs/custom-formio-components/providers/storage/s3custom.js
+++ b/app/frontend/components/shared/chefs/custom-formio-components/providers/storage/s3custom.js
@@ -1,31 +1,18 @@
 /* tslint:disable */
 const s3custom = function Provider(formio) {
   const requestPresignedUrl = (file, fileName, url) => {
-    return new Promise((resolve, reject) => {
-      const params = new URLSearchParams({
-        filename: fileName,
-        type: file.type,
-        size: file.size,
-      })
+    const params = new URLSearchParams({
+      filename: fileName,
+      type: file.type,
+      size: file.size,
+    })
 
-      const xhr = new XMLHttpRequest()
-      xhr.open("GET", url == "undefined" ? `/api/storage/s3?${params.toString()}` : `${url}?${params.toString()}`, true)
-      xhr.setRequestHeader("Content-Type", "application/json")
-      xhr.setRequestHeader("Accept", "application/json")
-
-      xhr.onreadystatechange = function () {
-        if (this.readyState === XMLHttpRequest.DONE) {
-          if (this.status === 200) {
-            const response = JSON.parse(this.responseText)
-            // console.log("***get signed url request", response)
-            resolve(response)
-          } else {
-            reject("Failed to get pre-signed URL")
-          }
-        }
-      }
-
-      xhr.send()
+    return fetch(url == "undefined" ? `/api/storage/s3?${params.toString()}` : `${url}?${params.toString()}`, {
+      method: "GET",
+      headers: {
+        "Content-Type": "application/json",
+        Accept: "application/json",
+      },
     })
   }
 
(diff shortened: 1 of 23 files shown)


app/frontend/components/shared/permit-applications/requirement-form.tsx
```diff
@@ -10,13 +9,14 @@ interface IRequirementFormProps {
...
-  const [form, setForm] = useState(permitApplication ? permitApplication.requirements : requirements)
-  // const [form, setForm] = useState(stubbedForm)
+  const form = permitApplication ? permitApplication.requirements : requirements
+  const [submission, setSubmission] = useState(permitApplication?.submissionData || {})
 
   //if there is no permit application provdided, you cannto run the key items like submit OR file upload.
   const onSubmit = (submission: any) => {
     if (permitApplication) {
       permitApplication.update({ submissionData: submission })
+      setSubmission(submission)
```
taw27, line 19: What would happen if there is an error with the submission

app/frontend/components/shared/permit-applications/requirement-form.tsx
```diff
@@ -10,13 +9,14 @@ interface IRequirementFormProps {
 }
 
 export function RequirementForm({ permitApplication, requirements }: IRequirementFormProps) {
-  const [form, setForm] = useState(permitApplication ? permitApplication.requirements : requirements)
-  // const [form, setForm] = useState(stubbedForm)
+  const form = permitApplication ? permitApplication.requirements : requirements
+  const [submission, setSubmission] = useState(permitApplication?.submissionData || {})
```
taw27, line 13: why is a separate state needed? mobx should be reactive if you change the function to observers

app/models/requirement.rb
```diff
@@ -94,30 +94,31 @@ def number_unit
...
+
     json
   end
 
   private
 
   # requirement codes should not be auto generated during seeding.  Use uuid if not provided
   def set_requirement_code
-    self.requirement_code ||= SecureRandom.uuid
+    self.requirement_code ||= label.present? ? label.parameterize.underscore.camelize(:lower) : SecureRandom.uuid
```
taw27, line 121: For this to work with the key method, we have to write a validation to ensure that the label is not the same as another requirement's label in the same requirement block

app/models/requirement_block.rb
```diff
@@ -27,11 +27,15 @@ def search_data
     }
   end
 
+  def key
+    "formSubmissionDataRequirementblock#{id}"
```
taw27, line 31: Will this cause an issue if a form template has different sections with same requirement block? Maybe have to move this to the join table function or pass in the section id in the to form json function from the sections

taw27 (changes requested):

app/models/requirement.rb
sunto: @taw27 - doe sit need to be unique? I guess the code needs to be unique per requirement block, not against all rewquirements.

app/models/requirement_block.rb
sunto: good pint, will add the section as well here.

app/models/requirement.rb
taw27: Never mind I misunderstood something

app/frontend/components/shared/permit-applications/requirement-form.tsx
sunto: One is for loading the 'previous' submission to restore saved state.  The setSubmission is being used when the file gets saved in the backend and the payload needs to get reformatted.  Will leave a nte maybe to see if this is working as expected.

taw27: @sunto but that should happen anyway if you update the submissionData on mobx, unless you are planning to do optimistic updates, this is not necessary. Even then you can just do this from within the action. Lemme know if you want to chat about this

app/frontend/components/shared/permit-applications/requirement-form.tsx
sunto: remvoed teh setSubmission.

app/frontend/components/shared/chefs/custom-formio-components/providers/storage/s3custom.js
```diff
@@ -115,28 +115,37 @@ const s3custom = function Provider(formio) {
...
+              const response = JSON.parse(this.responseText)
+              window.open(response.url, "_blank")
+              //NOT SURE WHY, BUT THIS RESOLVE DOES NOT CAUSE THE WINDOW.OPEN IN FORMIO
+              resolve(response)
+            } else {
+              reject("Failed to download file")
+            }
+          }
+        }
+        xhr.send()
```
taw27, line 146: Can't we do this using `fetch` instead of XMLHttpRequest

app/frontend/components/shared/chefs/custom-formio-components/providers/storage/s3custom.js
```diff
@@ -115,28 +115,37 @@ const s3custom = function Provider(formio) {
           .catch(reject)
       })
     },
-    deleteFile(fileInfo) {
-      console.log(fileInfo)
+    deleteFile(fileInfo, options) {
+      //assume we will not have public-read acl, use shrine to generate the request
+      console.log("s3 custom delete file", fileInfo)
```
taw27, line 120: remove console.log or wrap in import.meta.DEV

taw27 (approved):

PR #83, opened 2024-02-05
Feature/bphh 50/edit jurisdiction

bkmaibach:

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

re-seed needed. Log in as review_manager and go to content management


app/controllers/api/permit_applications_controller.rb
```diff
@@ -12,11 +12,6 @@ def show
   end
 
   def create
-    # authorize @permit_application
-    # render_success @permit_application, nil, { blueprint: PermitApplicationBlueprint }
-  end
-
-  def update
```
LaurelOlson, line 19: looks like the update route still exists - should remove that as well

PR #88, opened 2024-02-06
adding formio chefs styling

sunto:
## Description

Open a form from /permit-applications
You should see the submit button be dark blue like the BC gov colors

- [ x] 🍕 Feature


app/frontend/components/shared/chefs/styles.scss
```diff
@@ -0,0 +1,400 @@
...
+p {
+  color: $bcgov-font;
+  line-height: 1.7;
+}
+
+// links
+a {
+  color: $bcgov-link;
+  &:hover {
+    color: $bcgov-link-hover;
```
taw27, line 63: @sunto Github doesn't support scroll on comment selection so this is for line 25 to 154. I think these styles can be stripped out as they are for their application. I think we need only the formio overrides. It's also because the h1 and a tags etc is going to override our default styles

taw27: alternatively can wrap these inside the .form-wrapper class. That being said we should strip the styles that we don't need

taw27 (changes requested):

PR #97, opened 2024-02-08
Add parameterized classification option endpoint

bkmaibach:
## Description

Adds an endpoint that optionally shows only options that pertain to classifications of published templates.
Works for both classification types as well as the non-published case for seeing all available types, making it a universal endpoint for classification options.

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-539

## Steps to QA

Create a new template as superadmin, you should see all of the classification options right away
Create a new permit application as submitter. You should first see the published permit types, then after selecting you see the appropriate published work types.

### DIFF
--- a/app/frontend/services/api/index.ts
+++ b/app/frontend/services/api/index.ts
@@ -2,6 +2,7 @@ import { ApiResponse, ApisauceInstance, create, Monitor } from "apisauce"
 import { TCreateRequirementTemplateFormData } from "../../components/domains/requirement-template/new-requirement-tempate-screen"
 import { IJurisdiction } from "../../models/jurisdiction"
 import { IPermitApplication } from "../../models/permit-application"
+import { IPermitType } from "../../models/permit-classification"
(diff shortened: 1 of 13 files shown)


app/frontend/services/api/index.ts
```diff
@@ -105,6 +106,20 @@ export class Api {
...
+    type,
+    published = false,
+    permit_type_id: string = null,
+    activity_id: string = null
+  ) {
+    return this.client.post<IOptionResponse<IPermitType>>(`/permit_classifications/permit_classification_options`, {
+      type: type,
+      published: published,
+      permit_type_id: permit_type_id, // Use empty string if null
+      activity_id: activity_id, // Use empty string if null
```
taw27, line 119: What do the comments mean? pass in empty strings?

taw27 (approved):

PR #119, opened 2024-02-16
add permit application submission

bkmaibach:
## Description

Save form as draft and form submission

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-523

## Steps to QA

Log in as submitter
go to a permit application
enter some values
save with the button on the top right
submit by going to the bottom, clicking the sign button, and submit

### DIFF
--- a/app/frontend/components/domains/permit-application/successful-submission.tsx
+++ b/app/frontend/components/domains/permit-application/successful-submission.tsx
@@ -1,0 +1,23 @@
+import { Container, Heading } from "@chakra-ui/react"
+import { observer } from "mobx-react-lite"
+import React from "react"
+import { useTranslation } from "react-i18next"
+import { useMst } from "../../../setup/root"
+import { RouterLinkButton } from "../../shared/navigation/router-link-button"
+
+export const SuccessfulSubmissionScreen = observer(({}) => {
+  const { t } = useTranslation()
+  const {} = useMst()
+
+  return (
+    <Container>
+      <Heading>Submission success!</Heading>
+
+      <RouterLinkButton to={`/permit-applications`} variant="tertiary">
+        <Heading as="h3" fontSize="lg" color="text.link">
+          Permit Applications
+        </Heading>
+      </RouterLinkButton>
+    </Container>
+  )
+})
--- a/app/frontend/components/domains/permit-application/successful-submission.tsx
+++ b/app/frontend/components/domains/permit-application/successful-submission.tsx
@@ -4,9 +4,6 @@ import React from "react"
 import { RouterLinkButton } from "../../shared/navigation/router-link-button"
 
 export const SuccessfulSubmissionScreen = observer(() => {
-  // const { t } = useTranslation()
-  // const {} = useMst()
-
   return (
     <Container>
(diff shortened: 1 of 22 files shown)


app/policies/permit_application_policy.rb
```diff
@@ -26,7 +26,11 @@ def create?
   end
 
   def update?
-    record.submitter == user
+    record.draft? && record.submitter == user
```
sunto, line 29: we should add this to the specs in particular.  this is some funcitonality to prove that a submitted applicaiton cannot be updated.

app/frontend/components/domains/permit-application/successful-submission.tsx
```diff
@@ -0,0 +1,21 @@
+import { Container, Heading } from "@chakra-ui/react"
+import { observer } from "mobx-react-lite"
+import React from "react"
+import { RouterLinkButton } from "../../shared/navigation/router-link-button"
+
+export const SuccessfulSubmissionScreen = observer(() => {
+  // const { t } = useTranslation()
+  // const {} = useMst()
```
taw27, line 8: Remove comment for code cleanup

app/frontend/models/permit-application.ts
```diff
@@ -105,6 +105,16 @@ export const PermitApplicationModel = types
...
       return response
     }),
+
+    submit: flow(function* (params) {
+      const response = yield self.environment.api.submitPermitApplication(self.id, params)
+      if (response.ok) {
+        const { data: permitApplication } = response.data
+        self.rootStore.permitApplicationStore.mergeUpdate(permitApplication, "permitApplicationMap")
+      }
+      return response
```
taw27, line 115: I think this should be `response.ok`, otherwise the code below will always appear to be successful as the response object is always defined
`const onSubmit = async (submission: any) => {
    if (await permitApplication.submit({ submissionData: submission })) {
      navigate("/permit-applications/sucessful-submission")
    }
  }`

app/frontend/utils/index.ts
```diff
@@ -1,17 +1,18 @@
 import * as humps from "humps"
 import * as R from "ramda"
 
-const SUBMISSION_DATA_REGEX = /^formSubmissionData/
+const SUBMISSION_DATA_PREFIX = "formSubmissionData"
+const SECTION_PREFIX = "section"
```
taw27, line 5: `section` as the prefix will lead to issues if we have any other model property starting with `section`. This is very brittle as `section` is a generic name that could be used in other places and lead to unintentional bugs. We can add in for now, but I highly suggest looking for a work around for this

taw27 (changes requested):

app/frontend/models/permit-application.ts
bkmaibach: good catch!

app/frontend/utils/index.ts
bkmaibach: good point

taw27 (approved):

PR #139, opened 2024-02-27
add multi pid select

bkmaibach:
## Description

Pid endpoint is now pids endpoint. This provides options to an auto-popup selector that lets you choose the most relevant pid. If none are available, this can be skipped.

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-571

## Steps to QA

go to new permit
add an address
available pids should show up in a dropdown.

If the address has no associated pids, you should be able to skip pid selection

### DIFF
--- a/app/frontend/components/shared/select/selectors/sites-select.tsx
+++ b/app/frontend/components/shared/select/selectors/sites-select.tsx
@@ -1,25 +1,29 @@
-import { Input as ChakraInput, Flex, FormControl, FormLabel, HStack, InputGroup, Text } from "@chakra-ui/react"
+import { Flex, FormControl, FormLabel, HStack, InputGroup, Text } from "@chakra-ui/react"
 import { MapPin } from "@phosphor-icons/react"
 import { debounce } from "lodash"
 import { observer } from "mobx-react-lite"
-import React, { useCallback } from "react"
-import { useFormContext } from "react-hook-form"
+import * as R from "ramda"
+import React, { useCallback, useEffect, useRef, useState } from "react"
+import { Controller, useFormContext } from "react-hook-form"
 import { useTranslation } from "react-i18next"
-import { ControlProps, InputProps, OptionProps, components } from "react-select"
+import Select, { ControlProps, InputProps, OptionProps, components } from "react-select"
 import { useMst } from "../../../../setup/root"
 import { IOption } from "../../../../types/types"
 import { AsyncSelect, TAsyncSelectProps } from "../async-select"
 
 type TSitesSelectProps = {
-  fetchOptions: (query: string) => Promise<IOption[]>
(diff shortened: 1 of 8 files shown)


app/frontend/components/shared/select/selectors/sites-select.tsx
```diff
@@ -29,22 +33,31 @@ export const SitesSelect = observer(
...
+        selectControl.dispatchEvent(new MouseEvent("mousedown", { bubbles: true }))
       }
     }
 
     const debouncedFetchOptions = useCallback(debounce(fetchSiteOptions, 1000), [])
 
+    useEffect(() => {
+      setSiteSelected(!!pidWatch || !!(!fetchingPids && siteWatch && R.isEmpty(pidOptions)))
+    }, [pidWatch, siteWatch, pidOptions, fetchingPids])
+
```
taw27, line 60: Any reason to do this here instead of onChange?

app/frontend/components/shared/select/selectors/sites-select.tsx
taw27, line 1: Any reason we are using normal Select instead of AyncSelect here (see tags-select.tsx). There seems to be a mixture of react state and form state, and a handleChange function which seems to be not used anywhere. Not sure what the use case is, can you give me some examples what is the expected behaviour here so I can get better context?

app/services/wrappers/geocoder.rb
```diff
@@ -11,19 +11,20 @@ def default_headers
...
-          addressString: address_string,
-          locationDescriptor: "parcelPoint",
-          autoComplete: true,
-          brief: true,
-          maxResults: 10,
-        },
-      )
+  def site_options(address_string = nil, coordinates = nil)
+    site_params = {
+      # addressString: "525 Superior Street, Victoria, BC"
```
taw27, line 16: remove comment

app/frontend/components/shared/select/selectors/sites-select.tsx
bkmaibach: this needs to reset when the pid options change, when the isFetching status changes, etc

app/frontend/components/shared/select/selectors/sites-select.tsx
bkmaibach: AsyncSelect might be a way to do this also, but as far as I was aware that only gives a way to provide a function that is run once to populate the options, but how do you have it programatically re-fetch the options based on what happens somewhere else? handleChange is used within the "search addresses" input
once an address is selected, it then fetches the pids corresponding to that address.

taw27 (approved):

PR #141, opened 2024-02-27
update stubs for jurisdiction templates

bkmaibach:


bkmaibach:
This will require a more complete refactor, see https://hous-bssb.atlassian.net/browse/BPHH-702

PR #144, opened 2024-02-28
add editable nickname to permit applications and add submission infobox

bkmaibach:
## Description

add controller logic, migration, input and form, validation, etc for nickname
also add info box and readonly option for submitted forms

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-672
https://hous-bssb.atlassian.net/browse/BPHH-518

## Steps to QA

log in as submitter
view a draft permit, edit a nickname by clicking on, typing, clicking off
save and finish later - now goes back one to index after saving
go to a submitted application, see the blue info box
See fields are now disabled (readonly)

## UI Accessibility Checklist

passes wave

### DIFF
--- a/app/frontend/stores/permit-application-store.ts
+++ b/app/frontend/stores/permit-application-store.ts
@@ -17,7 +17,6 @@ export const PermitApplicationStoreModel = types
       permitApplicationMap: types.map(PermitApplicationModel),
       tablePermitApplications: types.array(types.reference(PermitApplicationModel)),
       currentPermitApplication: types.maybeNull(types.reference(PermitApplicationModel)),
-      isFetchingPermitApplications: types.optional(types.boolean, false),
     }),
     createSearchModel<EPermitApplicationSortFields>("searchPermitApplications")
   )
@@ -105,7 +104,6 @@ export const PermitApplicationStoreModel = types
           perPage: opts?.countPerPage ?? self.countPerPage,
         }
       )
-
       if (response.ok) {
         self.mergeUpdateAll(response.data.data, "permitApplicationMap")
         // dual purpose method also serves the submitters
--- a/app/models/concerns/traverse_data_json.rb
+++ b/app/models/concerns/traverse_data_json.rb
@@ -4,7 +4,9 @@ module TraverseDataJson
   #used to take a hash, deeply search and flatten the results into a file array
   #block expects an array to be returned
   def find_file_fields_and_transform!(hash, files, &block)
-    hash&.each do |key, value|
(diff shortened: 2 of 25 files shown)


app/frontend/stores/permit-application-store.ts
```diff
@@ -17,7 +17,6 @@ export const PermitApplicationStoreModel = types
       permitApplicationMap: types.map(PermitApplicationModel),
       tablePermitApplications: types.array(types.reference(PermitApplicationModel)),
       currentPermitApplication: types.maybeNull(types.reference(PermitApplicationModel)),
-      isFetchingPermitApplications: types.optional(types.boolean, false),
```
sunto, line 20: should we not do something in the UI rather than takin this out?

app/models/concerns/traverse_data_json.rb
```diff
@@ -4,7 +4,7 @@ module TraverseDataJson
   #used to take a hash, deeply search and flatten the results into a file array
   #block expects an array to be returned
   def find_file_fields_and_transform!(hash, files, &block)
-    hash.each do |key, value|
+    hash&.each do |key, value|
```
sunto, line 7: maybe we should put a guard clause on these instead?

app/models/permit_application.rb
```diff
@@ -55,6 +53,12 @@ def number_prefix
     jurisdiction.prefix
   end
 
+  def assign_default_nickname
+    nickname = "#{jurisdiction_name}: #{full_address || pid || pin || id}"
```
sunto, line 57: i think we can simplify this to 1 line, the return should happen automaticlaly.

self.nickname= "#{jurisdiction_name}: #{full_address || pid || pin || id}"

app/models/concerns/traverse_data_json.rb
sunto:
```
def find_file_fields_and_transform!(hash, files, &block)
 return {} if hash.nil?

app/frontend/stores/permit-application-store.ts
bkmaibach: this was left over from before the index used the search concern
now the createSearchModel provides a similar field isSearching, and so this is no longer necessary

app/models/concerns/traverse_data_json.rb
bkmaibach: sure

PR #149, opened 2024-03-01
Feature/bphh 727/nopid

bkmaibach:
Fix PID selector for multi pid, no pid
fix env example

Improve error messaging to do a basic puts for errors, render error now take a first paramter which is an error object or error array.

Now we can see errors on DEV and we can see erorrs in our console without having to set a pry

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,7 +20,7 @@ CHES_CLIENT_ID=""
 CHES_CLIENT_SECRET=""
 
 GEO_ALR_REST_URL=https://alr_rest_url_stub/arcserver/rest/services/
-GEO_LTSA_PARCELMAP_REST_URL=https://ltsa_parcel_map_url_stub/arcserver/rest/services/
+GEO_LTSA_PARCELMAP_REST_URL=
 
 
 # Geocoder variables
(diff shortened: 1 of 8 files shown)


sunto (approved):

app/controllers/concerns/base_controller_methods.rb
```diff
@@ -44,7 +44,13 @@ def render_success(resource, message_key = nil, opts = {})
   end
 
   # error_key maps to a translation key in en.yml
-  def render_error(error_key = nil, opts = {})
+  def render_error(error_or_errors, error_key = nil, opts = {})
+    if error_or_errors.respond_to?(:message)
+      puts error.message
```
sunto, line 49: Rails.logger.error error.message

app/controllers/concerns/base_controller_methods.rb
```diff
@@ -44,7 +44,13 @@ def render_success(resource, message_key = nil, opts = {})
...
+    elsif error_or_errors.respond_to?(:full_messages)
+      puts error.full_messages.join(", ")
```
sunto, line 51: Rails.logger.error error.full_messages.join(",")

PR #162, opened 2024-03-05
add download modal and zip supporting document logic

bkmaibach:
## Description

Download application individual files or all files as a zip

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-670
https://hous-bssb.atlassian.net/browse/BPHH-632


db/migrate/20240229235448_add_is_zip_to_supporting_documents.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsZipToSupportingDocuments < ActiveRecord::Migration[7.1]
+  def change
+    add_column :supporting_documents, :is_zip, :boolean
```
sunto, line 3: @bkmaibach - I think we should remove this one.  I don't think we should have the zip file against the supporting docs. 
The generated submission maybe, but not this.  It also should never be a zip.

PR #164, opened 2024-03-06
Initial Implementation for websockets, uses HA Redis for Openshift

derekyau:


config/initializers/anycable.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
```
LaurelOlson, line 1: Add new ENV variables to `.env.example`

LaurelOlson (approved):

PR #180, opened 2024-03-09
adding elective field rendering

sunto:
## Description

Seed some elective fields, make sure you publish them. 
Ensure the local jurisdiction modifies the electives and adds them and publishes it.

Open a permit applicaiton in jurisidction without an elective - verify
Open a permit applicaiton in jurisidction with an elective - verify

[BPHH-747](https://hous-bssb.atlassian.net/browse/BPHH-747)

### DIFF
--- a/app/frontend/utils/formio-component-traversal.ts
+++ b/app/frontend/utils/formio-component-traversal.ts
@@ -48,6 +48,21 @@ export const combineComplianceHints = (
   const blocksList = findPanelComponents(updatedJson.components)
   blocksList.forEach((panelComponent) => {
     if (blocksLookups[panelComponent.id]) {
+      if (blocksLookups[panelComponent.id]?.tip) {
+        panelComponent["tip"] = blocksLookups[panelComponent.id].tip
+      }
+      if (blocksLookups[panelComponent.id]?.["enabledElectiveFieldIds"]) {
+        panelComponent.components.forEach((subComp) => {
+          if (subComp.elective) {
+            subComp["conditional"] = {
+              show: blocksLookups[panelComponent.id]?.enabledElectiveFieldIds?.includes(subComp.id),
+              when: null,
+              eq: "",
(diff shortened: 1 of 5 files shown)


app/models/requirement.rb
```diff
@@ -141,6 +141,8 @@ def to_form_json(requirement_block_key = requirement_block&.key)
 
     json.merge!({ description: hint }) if hint
 
+    json.merge!({ elective: elective }) if elective
```
taw27, line 144: @sunto shouldn't we also add `show: false` by default if a requirement has `elective: true`?  Alternatively it can be done in `[formio-component-traversal.ts]`

app/frontend/utils/formio-component-traversal.ts
```diff
@@ -47,6 +47,21 @@ export const combineComplianceHints = (
...
+        panelComponent.components.forEach((subComp) => {
+          if (subComp.elective) {
+            subComp["conditional"] = {
+              show: blocksLookups[panelComponent.id]?.enabledElectiveFieldIds?.includes(subComp.id),
+              when: null,
+              eq: "",
+            }
+          }
+        })
+      }
```
taw27, line 63: if a jurisdiction does not make any customization this conditional will not pass `if (blocksLookups[panelComponent.id]?.["enabledElectiveFieldIds"])`. This is means if a field is marked elective, then we will display them by default which we not want. Remove the conditional and just have the following for it to work:

```
panelComponent.components.forEach((subComp) => {
          if (subComp.elective) {
            subComp["conditional"] = {
              show: blocksLookups[panelComponent.id]?.enabledElectiveFieldIds?.includes(subComp.id) ?? false,
              when: null,
              eq: "",
            }
          }
        })
```

taw27 (changes requested):

taw27 (approved):

PR #182, opened 2024-03-11
Bugfix/bphh 803/wrong inbox load

bkmaibach:
## Description

race condition was causing the jurisdiction to be set to null under some conditions
the fix uses the available jurisdiction ID when possible.

- [ ] 🍕 Feature
- [ x] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-803

## Steps to QA

log in as review manager
go to inbox
click home
click back
inbox should not load permit applications for null jurisdiction (ie submitter)


bkmaibach (commented):
small change so merging without review

PR #206, opened 2024-03-15
Feature/bphh 871/invite submitter

bkmaibach:
## Description

Provides a more descript error message when a user whose email is alreday taken is invited.

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-871

## Steps to QA

log in as review manager
try to invite any email that already exists (submitter@example.com)
error message should appear along with error tag


app/controllers/api/invitations_controller.rb
```diff
@@ -6,7 +6,7 @@ class Api::InvitationsController < Devise::InvitationsController
 
   def create
     inviter = Jurisdiction::UserInviter.new(inviter: current_user, users_params: users_params).call
-    if inviter.results[:invited].any?
+    if (inviter.results[:invited] + inviter.results[:email_taken]).any?
```
taw27, line 9: Just clarifying the logic. What does `inviter.results[:email_taken]` refer to? If it means that the email is taken, then should we still render_success?

bkmaibach: yes good question
we want to render success so that the UI can display the specific emails that were taken.
this was put in place because if for instance a single email was given, and it was taken, then before it would just show an error toast saying something failed. What I want is for the frontend to show exactly which emails were taken, and so this is deliberate.

taw27 (approved):

PR #210, opened 2024-03-18
updating computed compliance capabilities

sunto:
## Description

Automated compliance messages show up with text (PID, Plan Number, etc.)

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

(https://hous-bssb.atlassian.net/browse/BPHH-653)

## Steps to QA

Create a new application, it should set up PID and plan Number default values and show what the auto-compliance ofund as ad efault value.

## General Checklist

- [ x] 📝 Use descriptive commit messages
- [ x] 📗 Update any related documentation and include any relevant screenshots

### DIFF
--- a/app/models/concerns/automated_compliance_utils.rb
+++ b/app/models/concerns/automated_compliance_utils.rb
@@ -1,13 +1,9 @@
 module AutomatedComplianceUtils
   extend ActiveSupport::Concern
 
-  #placheholder, will be removed once lookup_props is refactored
-  def published_requirement_template
-    RequirementTemplate.find_by(activity: activity, permit_type: permit_type)
-  end
-
-  def requirements_lookups
-    published_requirement_template&.lookup_props || {}
+  def requirements_lookups #a flattened list of requiements with the right information
+    # template_version&.lookup_props
+    template_version&.requirement_template&.lookup_props
   end
 
   def automated_compliance_requirements
@@ -17,7 +13,6 @@ module AutomatedComplianceUtils
 
   def automated_compliance_unfilled_requirements
     automated_compliance_requirements.select do |field_id, req|
-      #TODO: if it is a file field, we check the compliance_data value is set instead
       submission_field_is_empty?(field_id, req)
     end
   end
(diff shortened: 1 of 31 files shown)


app/models/concerns/automated_compliance_utils.rb
```diff
@@ -1,13 +1,9 @@
...
-  #placheholder, will be removed once lookup_props is refactored
-  def published_requirement_template
-    RequirementTemplate.find_by(activity: activity, permit_type: permit_type)
-  end
-
-  def requirements_lookups
-    published_requirement_template&.lookup_props || {}
+  def requirements_lookups #a flattened list of requiements with the right information
+    # template_version&.lookup_props
+    template_version&.requirement_template&.lookup_props
```
taw27, line 6: should we rely on the requirement_template? The current draft `requirement_template` state might be vastly different from what's there in the currently published `template_version` e.g. Removed and added a bunch of requirements that isn't in the current template version

app/services/integrations/ltsa_parcel_map_bc.rb
```diff
@@ -26,7 +23,7 @@ def historic_site_by_pid(pid:)
     query = "where=HISTORIC_SITE_IND='Y' AND PARCEL_DESCRIPTION='#{pid}'&returnGeometry=true&outFields=*"
     response = @client.get("#{ENV["GEO_LTSA_PARCELMAP_REST_URL"]}#{HISTORIC_SERVICE}/query?f=json&#{query}")
     #assuem if there is a parcel description match not to use ltsa geometry matching
-    return response if response.success? && response.body.dig("features").length > 0
+    return parse_attributes_from_response(response) if response.success? && response.body.dig("features").length > 0
```
taw27, line 26: what happens if features is nil

app/models/concerns/automated_compliance_utils.rb
sunto: yes..I think we have separate card to remove thise lookup_props from the requirment template.  https://hous-bssb.atlassian.net/browse/BPHH-846

app/services/integrations/ltsa_parcel_map_bc.rb
sunto: if the pid fials to be found (returns nil), it goes onto look fo rthe geometry of the pid, then suppies that geometry for a intersection with historic sites to see if there is any overlap

taw27 (approved):

PR #259, opened 2024-03-22
Revert seed fix

LaurelOlson:
Note: in order for seeds to work, the Core Permit Requirements List xlsx file must be manually added to db/templates as this is no longer in source control.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -103,14 +103,11 @@ if PermitApplication.first.blank?
 
   puts "Seeding requirement templates..."
   # Create RequirementTemplate records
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -125,7 +122,7 @@
   # Creating Permit Applications
   puts "Seeding permit applications..."
   submitters = User.submitter
-  template_version = template1.published_template_version
+  template_version = RequirementTemplate.published.first
```
taw27, line 125: Screenshot 2024-03-22 at 4 27 06 PM
@seeds seem to be broken.

taw27 (changes requested):
Screenshot 2024-03-22 at 4 26 35 PM

@LaurelOlson

PR #294, opened 2024-03-27
add site configuration page, frontend and backend

bkmaibach:
## Description
site configuration (maintenance mode boolean and text)

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-801

## Steps to QA

log in as super admin
go to site configuration
set switch and text input
save and refresh
try logged out, different roles etc


karenkun:
Tested this and if I'm logged in as submitter, reviewer, reviewer_manager I won't see that new maintenance alert until I reload the page. Is that okay??

app/controllers/api/site_configuration_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def index
+    authorize :site_configuration, :index?
+    render_success @site_configuration
+  end
+
+  def create
+    authorize :site_configuration, :create?
+    @site_configuration.update(site_configuration_params)
+    render_success @site_configuration, "site_configuration.update_success"
```
taw27, line 15: no need to handle errors?

app/models/site_configuration.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def self.instance
+    first_or_create
+  end
+
+  private
+
+  # A private method to ensure only one record exists
+  def ensure_single_record
+    throw :abort if SiteConfiguration.count > 0
```
taw27, line 13: why not add normal validation errors.add ?

app/controllers/api/site_configuration_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Api::SiteConfigurationController < Api::ApplicationController
+  before_action :set_site_configuration, only: %i[index create]
+
+  skip_after_action :verify_policy_scoped
+  after_action :verify_authorized
```
taw27, line 5: isn't this only for index?

taw27 (changes requested):

app/controllers/api/site_configuration_controller.rb
bkmaibach: the reason these are here is to override the same in the parent, which specifies actions that dont exist in this subclass and cause errors

taw27 (approved):

PR #296, opened 2024-03-27
prevent permit full loads from closed download modal, emails from not…

bkmaibach:
… previously nil viewed_at

## Description

fix download modal from downloading full permit application thus setting viewed_at when its closed.

and only send the email when viewed_at was previously nil.

### DIFF
--- a/app/models/permit_application.rb
+++ b/app/models/permit_application.rb
@@ -209,10 +209,11 @@ class PermitApplication < ApplicationRecord
 
   def notify_user_application_updated
     return if new_record?
-
-    return unless saved_change_to_viewed_at? || saved_change_to_reference_number?
-
-    PermitHubMailer.notify_application_updated(submitter, self).deliver_later
+    viewed_at_change = previous_changes.dig("viewed_at")
+    # Check if the `viewed_at` was `nil` before the change and is now not `nil`.
+    if (viewed_at_change&.first.nil? && viewed_at_change&.last.present?) || saved_change_to_reference_number?
+      PermitHubMailer.notify_application_updated(submitter, self).deliver_later
+    end
   end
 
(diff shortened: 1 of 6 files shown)


app/models/permit_application.rb
```diff
@@ -209,10 +209,12 @@ def take_form_customizations_snapshot_if_submitted
 
   def notify_user_application_updated
     return if new_record?
-
-    return unless saved_change_to_viewed_at? || saved_change_to_reference_number?
-
-    PermitHubMailer.notify_application_updated(submitter, self).deliver_later
+    viewed_at_change = previous_changes.dig("viewed_at")
+    binding.pry
```
taw27, line 213: binding.pry

app/models/permit_application.rb
```diff
@@ -209,10 +209,12 @@ def take_form_customizations_snapshot_if_submitted
 
   def notify_user_application_updated
     return if new_record?
-
-    return unless saved_change_to_viewed_at? || saved_change_to_reference_number?
-
-    PermitHubMailer.notify_application_updated(submitter, self).deliver_later
+    viewed_at_change = previous_changes.dig("viewed_at")
```
taw27, line 212: Is it possible to use this api `attribute_changed?(attr_name)` , I believe it should give a boolean for if the attribute was changed

taw27: if not good to go

taw27 (changes requested):

app/models/permit_application.rb
bkmaibach: this is the only way I could get it to work. I'm trying to find not when it was changed but specifically when it was changed from nil

taw27 (approved):

PR #298, opened 2024-03-28
add action for changing role

bkmaibach:
## Description

Add role change to user table
Make sure to only allow review managers in the same jurisdiction
only allow changing to reviewer or review manager
had to reshuffle profile endpoint

log in as review manager, go to the users table, click manage, click change role, see it change!
be sure to test update of your own profile by going to the menu -> profile -> update name and save

https://hous-bssb.atlassian.net/browse/BPHH-912

### DIFF
--- a/app/frontend/models/user.ts
+++ b/app/frontend/models/user.ts
@@ -1,4 +1,4 @@
-import { Instance, flow, types } from "mobx-state-tree"
+import { Instance, applySnapshot, flow, types } from "mobx-state-tree"
 import { withEnvironment } from "../lib/with-environment"
 import { withRootStore } from "../lib/with-root-store"
 import { EUserRoles } from "../types/enums"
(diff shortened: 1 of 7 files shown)


app/frontend/models/user.ts
```diff
@@ -58,6 +58,18 @@ export const UserModel = types
       const response = yield self.environment.api.restoreUser(self.id)
       return response.ok
     }),
+    changeRole: flow(function* () {
+      if (self.role === EUserRoles.reviewManager) {
+        var newRole = EUserRoles.reviewer
+      } else if (self.role === EUserRoles.reviewer) {
+        var newRole = EUserRoles.reviewManager
```
taw27, line 65: change `var` to let and declare outside of if blocks, to keep consistent with modern javascript

app/frontend/models/user.ts
```diff
@@ -58,6 +58,18 @@ export const UserModel = types
...
+      } else {
+        return
+      }
+
+      const response = yield self.environment.api.updateUser(self.id, { role: newRole })
```
taw27, line 70: no need to update model `role`?

taw27 (approved):
I approved the changes, but added some clean up comments. One thing I think should be included is a confirmation modal before changing role

PR #323, opened 2024-04-03
support address site options with empty string values

bkmaibach:
## Description

Need to be able to select addresses with no site ID
this will trigger the mode for manually entering jurisdiciton PIN

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-970

## Steps to QA

log in as submitter, start a new application

use the address 5670 KULLAHUN => pick it from the dropdown => this should trigger the mode that displays PIN and jurisdiction selection
check another normal address

do the same thing on the jurisdiction search on the landing welcome page

### DIFF
--- a/app/services/wrappers/geocoder.rb
+++ b/app/services/wrappers/geocoder.rb
@@ -18,6 +18,12 @@ class Wrappers::Geocoder < Wrappers::Base
       brief: true,
       maxResults: 10,
       outputSRS: 4326,
+      # A few more params available for experimentation:
+      #   locationDescriptor: "any",
+      #   interpolation: "adaptive",
+      #   echo: true,
+      #   setBack: 0,
+      #   provinceCode: "BC"
     }
 
     site_params[:addressString] = address_string if address_string.present?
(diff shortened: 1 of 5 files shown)


app/services/wrappers/geocoder.rb
```diff
@@ -18,6 +18,12 @@ def site_options(address_string = nil, coordinates = nil)
       brief: true,
       maxResults: 10,
       outputSRS: 4326,
+      # A few more params available for experimentation:
+      #   locationDescriptor: "any",
+      #   interpolation: "adaptive",
+      #   echo: true,
+      #   setBack: 0,
+      #   provinceCode: "BC"
```
taw27, line 26: Comment intentional?

taw27 (approved):

app/services/wrappers/geocoder.rb
bkmaibach: yes

PR #346, opened 2024-04-05
add admin user search screen

bkmaibach:
## Description

refactor admin site config into sub-screens, including:
admin user search. This moves the table users into the user store instead of per-jurisdiciton.
Tests were added to ensure no unauthorized access. bonus: added search specs for permit applications.

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1037

## Steps to QA

Log in as super admin
go to site configuration
try the sitewide message and the admin user search.

### DIFF
--- a/app/frontend/components/domains/super-admin/site-configuration-management/users-screen.tsx
+++ b/app/frontend/components/domains/super-admin/site-configuration-management/users-screen.tsx
@@ -1,16 +1,13 @@
 import { Box, Container, Flex, Heading, VStack } from "@chakra-ui/react"
 import { format } from "date-fns"
 import { observer } from "mobx-react-lite"
-import React from "react"
+import React, { useEffect } from "react"
 import { useTranslation } from "react-i18next"
-import { useJurisdiction } from "../../../../hooks/resources/use-jurisdiction"
 import { useSearch } from "../../../../hooks/use-search"
 import { IUser } from "../../../../models/user"
 import { useMst } from "../../../../setup/root"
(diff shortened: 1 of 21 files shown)


app/frontend/components/domains/super-admin/site-configuration-management/users-screen.tsx
taw27, line 1: if it's a single component file, can we rename the file name to match the component name?

taw27 (approved):

spec/concerns/admin_users_search_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+    end
+
+    context "when attempting the search as a review_manager user role" do
+      let(:cur_user) { create(:user, :review_manager, jurisdiction: jurisdiction) }
+      let(:user_search_params) { { query: "", page: 1, per_page: 10 } }
+
+      it "returns all jurisdiction users for review_manager" do
+        controller.perform_user_search
+        expect(controller.instance_variable_get(:@user_search).results).to be_empty
```
LaurelOlson, line 61: this doesn't seem to match the conditional

LaurelOlson (commented):
This might be unrelated but when testing this, noticed the "Change role" option under manage on the users table doesn't do anything. From the code it looks like this is only relevant when viewing reviewers / review managers so it probably shouldn't show up if the user is an admin?

PR #371, opened 2024-04-11
fix permit and profile update issues, email confrim redirect, and copies

bkmaibach:
## Description

change redirect urls on email change confirm
put contact edit info outside of button
fix nickname update
update invite expired copy


config/routes.rb
```diff
@@ -121,6 +121,7 @@
 
   root to: "home#index"
 
+  get "/" => "home#index", :as => :home
```
derekyau, line 124: home#index can already be achieved by using the `root_url` defined above this line @bkmaibach

app/controllers/api/confirmations_controller.rb
```diff
@@ -19,9 +19,9 @@ def show
 
     if resource.errors.empty?
       PermitHubMailer.welcome(@user).deliver_later
-      redirect_to login_url(frontend_flash_message("user.confirmation_success", "success"))
+      redirect_to confirmed_url(frontend_flash_message("user.confirmation_success", "success"))
     else
-      redirect_to login_url(
+      redirect_to home_url(
```
derekyau, line 24: use `root_url`

PR #400, opened 2024-04-25
refactor integrations to use wrapper base class

bkmaibach:
## Description

refactor integrations to use wrapper base class
This consolidates error handling and parsing logic, and cleans up the use or faraday

https://hous-bssb.atlassian.net/browse/BPHH-703

Lots of rspec failures, but all fixed!
No changes to specs or cassettes besides class rename


app/services/wrappers/base.rb
```diff
@@ -13,19 +13,61 @@ def client
...
+    skip_handle_response ? response : handle_response(response)
+  rescue Faraday::Error => e
+    handle_faraday_error(e)
   end
 
-  def delete(path)
-    JSON.parse(client.delete(path).env.response_body)
+  def put(path, payload = {}, skip_handle_response = false)
+    response = client.put(path, payload)
+    skip_handle_response ? response : handle_response(response)
```
taw27, line 40: I see this ternary several times in this file. Might be a bit cleaner rename `skip_handle_response` to some other property like `skip_response_processing` and handle this flag inside the `handle_response`. This way you just have perform this logic at one place

app/services/wrappers/geocoder.rb
```diff
@@ -29,7 +27,7 @@ def site_options(address_string = nil, coordinates = nil)
     site_params[:addressString] = address_string if address_string.present?
     site_params[:parcelPoint] = coordinates.join(",") if coordinates.present?
 
-    r = get("/addresses.#{OUTPUT_FORMAT}", site_params)
```
taw27, line 32: I don't see a reason the output format variable needs to refactored?

app/services/wrappers/ltsa_parcel_map_bc.rb
```diff
@@ -0,0 +1,158 @@
...
+    if response.success?
+      return parse_attributes_from_response(response)
+    else
+      raise Errors::FeatureAttributesRetrievalError
+    end
+  end
+
+  def get_coordinates_by_pid(pid)
+    response = get_details_by_pid(pid: pid)
+    if response.success?
```
taw27, line 131: if the base `get` method already raises an error's for non 200 responses wouldn't the else condition never execute?

taw27 (approved):

app/services/wrappers/geocoder.rb
bkmaibach: its just unnecessary as we have established will always use the JSON output format. Should I put this back?

app/services/wrappers/ltsa_parcel_map_bc.rb
bkmaibach: the base offers the ability to skip handling the response, as to accommodate the way Sunny set up his integrations class. skip handle response is passed as true when more detailed control over the response is needed.

PR #447, opened 2024-05-13
Feature/bphh 1234/version diff

bkmaibach:
## Description

Show a clickable changes box to review manager on the permit template when navigating from the "changes on template" notification.

ticket:
https://hous-bssb.atlassian.net/browse/BPHH-1234

## Steps to QA:

See the card. Add and remove requirements as super admin, save, log in as review manager, navigate using hte notificaiton, see the changes in the floating sidebar. Clicking links should navigate to the appropriate block if it is still present.

### DIFF
--- a/app/frontend/components/domains/requirement-template/screens/jurisdiction-edit-digital-permit-screen/index.tsx
+++ b/app/frontend/components/domains/requirement-template/screens/jurisdiction-edit-digital-permit-screen/index.tsx
@@ -4,15 +4,24 @@ import { observer } from "mobx-react-lite"
 import React, { useEffect, useState } from "react"
 import { useForm } from "react-hook-form"
 import { useTranslation } from "react-i18next"
-import { useNavigate } from "react-router-dom"
+import { useLocation, useNavigate } from "react-router-dom"
 import { useJurisdictionTemplateVersionCustomization } from "../../../../../hooks/resources/use-jurisdiction-template-version-customization"
 import { useTemplateVersion } from "../../../../../hooks/resources/use-template-version"
 import { IJurisdictionTemplateVersionCustomization } from "../../../../../models/jurisdiction-template-version-customization"
+import { IRequirement } from "../../../../../models/requirement"
 import { useMst } from "../../../../../setup/root"
-import { IRequirementBlockCustomization, ITemplateCustomization } from "../../../../../types/types"
+import { ERequirementChangeAction } from "../../../../../types/enums"
+import {
+  ICompareRequirementsBoxData,
+  IRequirementBlockCustomization,
+  ITemplateCustomization,
+  ITemplateVersionDiff,
+} from "../../../../../types/types"
 import { ErrorScreen } from "../../../../shared/base/error-screen"
 import { LoadingScreen } from "../../../../shared/base/loading-screen"
+import { SharedSpinner } from "../../../../shared/base/shared-spinner"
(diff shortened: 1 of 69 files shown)


app/frontend/components/domains/requirement-template/screens/jurisdiction-edit-digital-permit-screen/index.tsx
```diff
@@ -83,6 +92,36 @@ export const JurisdictionEditDigitalPermitScreen = observer(function Jurisdictio
     reset(formFormDefaults(jurisdictionTemplateVersionCustomization))
   }, [jurisdictionTemplateVersionCustomization?.customizations])
 
+  const location = useLocation()
+  const queryParams = new URLSearchParams(location.search)
+  const isCompare = queryParams.get("compare") === "true"
```
taw27, line 97: since we are using react router, please use their apis when possible https://reactrouter.com/en/main/hooks/use-search-params

app/frontend/components/shared/permit-applications/scroll-link.tsx
```diff
@@ -1,16 +1,39 @@
...
+
+      // Focus on the target form control if present
+      const id = targetElement.querySelector("label.col-form-label")?.getAttribute("for")
+      const formControlElement = id ? document.getElementById(id) : null
       formControlElement?.focus?.()
+
+      targetElement.addEventListener("blur", () => {
+        targetElement.removeAttribute("tabIndex") // Removes the tabIndex attribute, returning it to its default behavior
+      })
     }
```
taw27, line 37: This will create a lot of event listeners as for every click a new listener is attached. You will have to look into it, but find a way also cleanup listeners as this might cause memory issues. I quickly chat gpted it https://chatgpt.com/share/811f6752-d68f-4b8e-9df4-c3480bc00fdf, and the solution seems simple enough but I did not look too deeply into it

app/models/template_version.rb
```diff
@@ -43,6 +45,76 @@ def form_json_requirements
...
+    .template_versions
+    .where(
+      "version_date <=? AND status = ? AND deprecation_reason = ?",
+      template_version.version_date,
+      TemplateVersion.statuses[:deprecated],
+      TemplateVersion.deprecation_reasons[:new_publish],
+    )
+    .where.not(id: template_version.id)
+    .order(version_date: :desc, created_at: :desc)
+    .first
```
taw27, line 70: why not just reuse the following in TemplateVersioningService.

```
 def self.previous_published_version(template_version)
    template_version
      .requirement_template
      .template_versions
      .where(
        "version_date <=? AND status = ? AND deprecation_reason =
 ?",
        template_version.version_date,
        TemplateVersion.statuses[:deprecated],
        TemplateVersion.deprecation_reasons[:new_publish],
      )
      .where.not(id: template_version.id)
      .order(version_date: :desc, created_at: :desc)
      .first
  end
```
Screenshot 2024-05-28 at 11 19 38 AM

Just make the class method public

taw27: The point of the service is to abstract away some of this logic into the service. You can call the service from the this model method

taw27 (changes requested):

app/models/template_version.rb
bkmaibach: ok

taw27 (approved):

PR #451, opened 2024-05-14
add help drawer content and manage screen

bkmaibach:
## Description

add management and sidebar for help links

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1427

## Steps to QA

migrate - this will add new default content
log in as super admin
go to site management > help links
add some HREFs - make sure they start with https://
save

go to nav menu > help to see the links

### DIFF
--- a/app/frontend/components/shared/form/input-form-control.tsx
+++ b/app/frontend/components/shared/form/input-form-control.tsx
@@ -41,7 +41,7 @@ export const TextFormControl = (props: IInputFormControlProps) => {
           inputProps: { type: "text" },
           validate: {
             satisfiesLength: (str) =>
-              ((!props.required || str?.length >= 1) && str?.length < 128) || t("ui.invalidInput"),
+              (!props.required && !str) || (str?.length >= 1 && str?.length < 128) || t("ui.invalidInput"),
           },
         },
         props
@@ -79,7 +79,7 @@ export const TextAreaFormControl = (props: IInputFormControlProps) => {
           inputProps: { as: Textarea }, // Use Chakra's Textarea instead of Input
           validate: {
             satisfiesLength: (str) =>
-              ((!props.required || str?.length >= 1) && str?.length < 512) || "Input is invalid", // Use a default error message or use translation as needed
+              (!props.required && !str) || (str?.length >= 1 && str?.length < 512) || "Input is invalid", // Use a default error message or use translation as needed
           },
         },
         props
--- a/app/frontend/components/shared/form/input-form-control.tsx
+++ b/app/frontend/components/shared/form/input-form-control.tsx
@@ -10,6 +10,7 @@ import {
   InputGroup,
   InputLeftElement,
   InputRightElement,
+  Textarea,
 } from "@chakra-ui/react"
 import { AsteriskSimple, Phone } from "@phosphor-icons/react"
 import { t } from "i18next"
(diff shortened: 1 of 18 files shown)


app/frontend/components/shared/form/input-form-control.tsx
```diff
@@ -70,6 +71,22 @@ export const FileFormControl = (props: IInputFormControlProps) => {
...
 }
 
+export const TextAreaFormControl = (props: IInputFormControlProps) => {
+  return (
+    <InputFormControl
+      {...(R.mergeDeepRight(
+        {
+          inputProps: { as: Textarea }, // Use Chakra's Textarea instead of Input
+          validate: {
+            satisfiesLength: (str) => !props.required || (str?.length >= 1 && str?.length < 256) || "Input is invalid", // Use a default error message or use translation as needed
```
taw27, line 81: shouldn't this `satisfiesLength: (str) => (!props.required && !value?.trim() || (str?.length >= 1 && str?.length < 256) || "Input is invalid",` . We still want to validate the length if they enter something even if the input value is not required

app/frontend/components/shared/help-drawer.tsx
```diff
@@ -44,22 +57,85 @@ export function HelpDrawer({ defaultButtonProps, renderTriggerButton }: IProps)
...
-                h: "full",
-              },
-            }}
-          >
-            <iframe src="https://www2.gov.bc.ca/gov/content?id=A5A88A4CE1D54D95AB23D57858EF11EE" title="Help"></iframe>
+          <DrawerBody>
+            {helpLinkItems && (
+              <Flex direction="column" gap={6}>
+                {Object.keys(helpLinkItems)
+                  .filter((key) => helpLinkItems[key].show)
```
taw27, line 67: might better to move this logic to a `view` in the MST model as that should be a bit more performant

app/frontend/components/shared/help-drawer.tsx
```diff
@@ -44,22 +57,85 @@ export function HelpDrawer({ defaultButtonProps, renderTriggerButton }: IProps)
...
+            <Box w={6}>{icon}</Box>
+            <Heading as="h3" mb={0} fontSize="lg">
+              {title}
+            </Heading>
+          </HStack>
+          <Text ml={8} fontSize="sm">
+            {description}
+          </Text>
+        </Flex>
+        <Link href="https://www.google.com" isExternal>
```
taw27, line 131: is this supposed to go to google.com?

app/models/site_configuration.rb
```diff
@@ -10,6 +13,34 @@ def self.instance
...
+        end
+      end
+
+      # Check if href is a valid URL
+      if item["href"].present?
+        unless item["href"].match?(%r{\Ahttps://.+\z})
+          errors.add(
+            :base,
+            I18n.t("activerecord.errors.models.site_configuration.attributes.help_link_items.invalid_url", link: key),
+          )
```
taw27, line 41: can you follow the same method as this https://github.com/bcgov/HOUS-permit-portal/blob/develop/app/models/concerns/validate_url_attributes.rb . You can limit to https, but prefer this over regex

taw27 (changes requested):

app/frontend/components/shared/form/input-form-control.tsx
```diff
@@ -78,7 +78,7 @@ export const TextAreaFormControl = (props: IInputFormControlProps) => {
         {
           inputProps: { as: Textarea }, // Use Chakra's Textarea instead of Input
           validate: {
-            satisfiesLength: (str) => !props.required || (str?.length >= 1 && str?.length < 256) || "Input is invalid", // Use a default error message or use translation as needed
+            satisfiesLength: (str) => (str?.length >= 1 && str?.length < 256) || "Input is invalid", // Use a default error message or use translation as needed
```
taw27, line 81: what happens if the value is an empty string and the property is not required? Expected behaviour is the input to be valid for blank value, but if there is a value present then it should validate it's length. Can you double check if this condition works?

taw27 (approved):

app/frontend/components/shared/form/input-form-control.tsx
```diff
@@ -78,7 +78,8 @@ export const TextAreaFormControl = (props: IInputFormControlProps) => {
         {
           inputProps: { as: Textarea }, // Use Chakra's Textarea instead of Input
           validate: {
-            satisfiesLength: (str) => !props.required || (str?.length >= 1 && str?.length < 256) || "Input is invalid", // Use a default error message or use translation as needed
+            satisfiesLength: (str) =>
+              ((!props.required || str?.length >= 1) && str?.length < 512) || "Input is invalid", // Use a default error message or use translation as needed
```
taw27, line 82: @bkmaibach I think the logic should be `(!props.required && !str) || (str?.length >= 1 && str?.length < 512) || t("ui.invalidInput")`. The same thing on the one above

taw27 (approved):

PR #471, opened 2024-05-27
add last minute landing ui updates

bkmaibach:
Change the section with download buttons on landing

Add a "These communities are enabled" info box

### DIFF
--- a/app/frontend/i18n/i18n.ts
+++ b/app/frontend/i18n/i18n.ts
@@ -119,11 +119,18 @@ const options = {
           goTo: "Go to {{ location }}",
           permitApp: "permit app",
           adminPanel: "admin panel",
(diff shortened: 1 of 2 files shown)


app/frontend/i18n/i18n.ts
```diff
@@ -119,11 +119,18 @@ const options = {
           goTo: "Go to {{ location }}",
           permitApp: "permit app",
           adminPanel: "admin panel",
+          enabledCommunitiesDescription: "Communities you can submit Building Permit applications in:",
+          enabledCommunities: [
+            { name: "City of North Vancouver", href: "/jurisdictions/corporation-of-the-city-of-north-vancouver" },
+          ],
```
taw27, line 125: can you move this to constants. Seems awkward to have data structures in the translation file. You can place the name translation here.

taw27 (changes requested):

PR #474, opened 2024-05-28
Integration/notification

bkmaibach:
## Description

Integration PR for all notification/compare code


app/models/template_section_block.rb
```diff
@@ -4,31 +4,4 @@ class TemplateSectionBlock < ApplicationRecord
...
-
-  def duplicate_requirement_block_on_template_exists?
-    requirement_template
-      .requirement_template_sections
-      .joins(:template_section_blocks)
-      .where(template_section_blocks: { requirement_block: requirement_block })
-      .where.not(template_section_blocks: { id: id })
-      .limit(1)
-      .exists?
-  end
```
taw27, line 33: Is this validation replicated somewhere else, if not then is this constraint not required anymore?

taw27 (approved):

app/models/template_section_block.rb
bkmaibach: I'm not sure why this was removed, I'll put it back

bkmaibach: nevermind, this was moved on purpose to requirement_template.rb in 
https://github.com/bcgov/HOUS-permit-portal/commit/27ce318359d11c04444f59509a29e76c090d1d36

PR #480, opened 2024-05-30
add backend search data for if a jurisdiciton is set up and show on landing

bkmaibach:
## Description

determining if an inbox is set up for a jurisdiction is now determined on the backend and indexed in search data.
Add this search data to the jurisdiction tables and show up to 10 of the enabled jurisdictions on the landing.

@Skye Smith wants us to say “not accepting submissions” until ALL inboxes are set up (this means that the message will not be accurate for some permit types)

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1496

## Steps to QA

see card

### DIFF
--- a/app/frontend/stores/jurisdiction-store.ts
+++ b/app/frontend/stores/jurisdiction-store.ts
@@ -122,8 +122,8 @@ export const JurisdictionStoreModel = types
     },
   }))
   .actions((self) => ({
-    searchEnabledJurisdictions: flow(function* () {
-      return self.searchJurisdictions({ reset: true, page: 1, countPerPage: 12 }, true)
+    searchEnabledJurisdictions: flow(function* (countPerPage: number = 10) {
+      return self.searchJurisdictions({ reset: true, page: 1, countPerPage }, true)
     }),
   }))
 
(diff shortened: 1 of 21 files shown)


app/frontend/stores/jurisdiction-store.ts
```diff
@@ -119,5 +121,10 @@ export const JurisdictionStoreModel = types
       return j?.id
     },
   }))
+  .actions((self) => ({
+    searchEnabledJurisdictions: flow(function* () {
+      return self.searchJurisdictions({ reset: true, page: 1, countPerPage: 12 }, true)
```
taw27, line 126: Just confirming hard coded values are on purpose and not for just testing

app/models/jurisdiction.rb
```diff
@@ -127,6 +128,14 @@ def unviewed_permit_applications
     permit_applications.unviewed
   end
 
+  def submission_inbox_set_up
+    PermitType.all.all? do |permit_type|
+      self.permit_type_submission_contacts.any? do |contact|
+        contact.permit_type_id == permit_type.id && contact.email.present? && contact.confirmed_at.present?
+      end
+    end
+  end
```
taw27, line 137: can you enter this in chat GPT. I suspect there is a way to just do this in a single query using joins rather than looping through them all. Just a suggestion to explore

taw27 (approved):

PR #490, opened 2024-06-06
Add revision mode and buttons

bkmaibach:
## Description

This adds a revision mode and formjson mutation for adding in buttons, with the necessary CSS workarounds.

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1455

## Steps to QA

see card

### DIFF
--- a/app/frontend/utils/formio-component-traversal.ts
+++ b/app/frontend/utils/formio-component-traversal.ts
@@ -158,24 +158,24 @@ export const combineDiff = (formJson: IFormJson, diff: ITemplateVersionDiff) => 
   return formJson
 }
 
-export const combineRevisionButtons = (formJson: IFormJson) => {
-  const convertToRevisionButton = (requirement: IFormIORequirement) => {
-    return {
-      id: requirement.id + "-revision-button",
-      key: requirement.key + "-revision-button",
-      type: "button",
-      label: "",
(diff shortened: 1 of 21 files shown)


app/frontend/utils/formio-component-traversal.ts
```diff
@@ -156,3 +157,41 @@ export const combineDiff = (formJson: IFormJson, diff: ITemplateVersionDiff) =>
...
+      title: "Revision Button",
+      input: true,
+      action: "custom",
+      custom: `console.log("${requirement.key} says hello")`,
+      customClass: "revision-button",
+      hideLabel: true,
+      customConditional: requirement.customConditional,
+      conditional: requirement.conditional,
+    } as IFormIORequirement
+  }
```
taw27, line 177: I don't see a need for this to be in side the parent function as it's not referencing anything internally. Place it outside so that a new function is not created every time the parent function is called

taw27 (approved):

app/frontend/utils/formio-component-traversal.ts
bkmaibach: good idea

PR #492, opened 2024-06-07
Feature/bphh 1510/jurisdiction field map api

taw27:
## Description
- Add back-end model and data migration to add `JurisdictionIntegrationRequirementsMapping`
- Adds backend call back logic to:
    - Initializes mapping and copies over existing mappings from other templates when a new `JurisdictionIntegrationRequirementsMapping` instance is created
    - Creates mappings for any template version which already does not have a mapping, when api is enabled in jurisdiction
    - Creates missing mapping for any Jurisdiction with api enabled for the template version when the template version in published
    - Syncs common mappings between templates of the same jurisdiction
 - Adds `GET {base_rote}/template_versions/:id/jurisdictions/:jurisdiction_id/jurisdiction_integration_requirements_mapping` and `PATCH {base_rote}/jurisdiction_integration_requirements_mapping`

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1510
https://hous-bssb.atlassian.net/browse/BPHH-1511

## Steps to QA
- Run `rails db:migrate:with_data`
- Ensure there is a published template version
- Update any Jurisdiction to enable the api. `@jurisdiction.update(external_api_enabled: true`
- Run `bundle exec rspec` to verify call back specs are passing
- You should be able to make the following requests via post man now:
Screenshot 2024-06-06 at 3 06 01 PM

Screenshot 2024-06-06 at 4 55 26 PM

### DIFF
--- a/app/controllers/api/jurisdiction_integration_requirements_mappings_controller.rb
+++ b/app/controllers/api/jurisdiction_integration_requirements_mappings_controller.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class Api::JurisdictionIntegrationRequirementsMappingsController < Api::ApplicationController
+  before_action :set_jurisdiction_integration_requirements_mapping, only: %i[update]
+
+  def update
+    authorize @jurisdiction_integration_requirements_mapping
+
+    if @jurisdiction_integration_requirements_mapping.update_requirements_mapping(
+         jurisdiction_integration_requirements_mapping_params.to_h[:simplified_map],
+       )
+      render_success @jurisdiction_integration_requirements_mapping,
+                     "jurisdiction_integration_requirements_mapping.update_success"
+    else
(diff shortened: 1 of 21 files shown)


bkmaibach:
Sometimes I like to do this:

image

app/controllers/api/jurisdiction_integration_requirements_mappings_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class Api::JurisdictionIntegrationRequirementsMappingsController < Api::ApplicationController
```
bkmaibach, line 3: These class names are really giving the ultrawide a run for its money

bkmaibach (approved):

PR #500, opened 2024-06-13
filter out of date permit applications for notification link

bkmaibach:
## Description

Refactor search filters in order to provide a link for template update notifications 

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1612

## Steps to QA

Update template as super admin
get notifcation as submitter
click the Go link
You should see only the relevant draft permits with yellow highlighting
reset the filter to see everything again.

### DIFF
--- a/app/frontend/lib/create-search-model.ts
+++ b/app/frontend/lib/create-search-model.ts
@@ -1,5 +1,5 @@
 import { flow, Instance, types } from "mobx-state-tree"
-import { EPermitApplicationStatus, ESortDirection } from "../types/enums"
+import { ESortDirection } from "../types/enums"
 import { ISort } from "../types/types"
 import { setQueryParam } from "../utils/utility-functions"
 
@@ -9,11 +9,9 @@ interface IFetchOptions {
   countPerPage?: number
 }
 
-const filterableStatuses = Object.values(EPermitApplicationStatus)
-export type TFilterableStatus = (typeof filterableStatuses)[number]
-
 export const createSearchModel = <TSortField, TFetchOptions extends IFetchOptions = IFetchOptions>(
-  fetchDataActionName: string
+  fetchDataActionName: string,
+  setFiltersName: string
 ) =>
   types
     .model()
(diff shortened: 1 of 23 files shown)


app/frontend/lib/create-search-model.ts
```diff
@@ -9,11 +9,9 @@ interface IFetchOptions {
   countPerPage?: number
 }
 
-const filterableStatuses = Object.values(EPermitApplicationStatus)
-export type TFilterableStatus = (typeof filterableStatuses)[number]
-
 export const createSearchModel = <TSortField, TFetchOptions extends IFetchOptions = IFetchOptions>(
-  fetchDataActionName: string
+  fetchDataActionName: string,
+  setFiltersName: string
```
taw27, line 14: make the second field optional as not all search models need filters.

app/frontend/stores/requirement-block-store.ts
```diff
@@ -123,6 +123,10 @@ export const RequirementBlockStoreModel = types
       }
       return response.ok
     }),
+    setRequirementBlockFilters(queryParams) {
+      // Set filters here once necessary
+      // See permit-application-store.ts for example
+    },
```
taw27, line 129: idea the create searchModel should handle this if possible. Just make the filters param optional in the create function and escape filter related logic based on that. As not all searches will have filters

app/frontend/stores/requirement-template-store.ts
```diff
@@ -79,6 +79,10 @@ export const RequirementTemplateStoreModel = types
       }
       return response.ok
     }),
+    setRequirementTemplateFilters(queryParams) {
+      // Set filters here once necessary
+      // See permit-application-store.ts for example
+    },
```
taw27, line 85: same comment as before

app/frontend/stores/user-store.ts
```diff
@@ -142,6 +142,10 @@ export const UserStoreModel = types
       }
       return response.ok
     }),
+    setUserFilters(queryParams) {
+      // Set filters here once necessary
+      // See permit-application-store.ts for example
+    },
```
taw27, line 148: same comment as before

taw27 (changes requested):

app/frontend/stores/requirement-block-store.ts
bkmaibach: I'm not sure I understand, it does handle this (in the same way it handles fethData using fetchDataActionName)

taw27: @bkmaibach I am saying this shouldn't be required as not all searches use filtering

taw27 (approved):

PR #506, opened 2024-06-14
add migration to force publish all templates

bkmaibach:
## Description

Required to get everything depending on template version data to display the updated permit type names


taw27 (changes requested):
This is too aggressive as it will override all current templates with what ever is in draft. 

This is fine for dev, but I not sure if it's ok for test and prod, which this will eventually go in, so please connect with Skye for that. 

Depending on the above, please connect with me on Monday, to browse alternatives, unless it's immediately urgent.

Also for future references, we are moving to use this gem https://github.com/ilyakatz/data-migrate for all data migrations as per my conversation with @derekyau previously.

db/data/20240617164456_update_template_version_permit_types.rb
```diff
@@ -0,0 +1,51 @@
...
+    PermitType.find_by_code("medium_residential").update(
+      name: "Medium density Residential",
+      description: "Part 9 townhouses, small apartment buildings",
+    )
+    PermitType.find_by_code("high_residential").update(
+      name: "High density residential",
+      description: "Highest density residential structures",
+    )
+
+    update_template_version_jsons
```
taw27, line 35: I think the down is mistakenly copied over from up. keep it blank is there is no need for down

taw27 (approved):

PR #508, opened 2024-06-17
Fix issues re-inviting users

LaurelOlson:
## Description

Fixes issues inviting a user that already has a submitter account.

Previously, if a review staff member was invited but did not accept the
invite and instead created a submitter account, they could not always be
re-invited as a review staff. This occurred if the submitter was
retrieved first when inviting by email. In this scenario, the submitter
would be updated to have the same nickname as the existing review staff.
This triggered an error since the nickname field has a unique
constraint. This exposed the real error, which is that the submitter
should not be considered when retrieving the user to invite.

This also fixes a related issue where a submitter is invited as a review
staff for the first time. Instead of sending an invite to the review
staff as a new user, the submitter was automatically converted to a
review staff without needing to accept the invite.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

[BPHH-1636](https://hous-bssb.atlassian.net/browse/BPHH-1636)

## Steps to QA

1. Login as a super admin or review manager and invite a reviewer or review manager. Note the email.
2. In a separate session, create a submitter account. Confirm the profile, setting the notification email to the email that was used in the invitation in step 1.
3. As the super admin or review manager, try to re-invite the invited user from step 1. The invitation should send.
4. Open the invitation in a separate session and sign in with the same BCeID used in step 2. 

After accepting the invitation, the user should be prompted to accept the EULA and then be logged in as the invited user type (reviewer or review manager). 

## UI Accessibility Checklist

- [ ] Checked
      for accessibility errors?

## General Checklist

- [x] ✅ Provide tests for your changes where relevant

N/A


db/schema.rb
```diff
@@ -75,24 +75,6 @@
     t.integer "variant"
   end
 
-  create_table "external_api_keys",
```
sunto, line 78: this seems odd. I think we need to make sure these removals on api is not checked in.

LaurelOlson: This was accidentally committed onto `main` from `develop` [here](https://github.com/bcgov/HOUS-permit-portal/pull/458/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3) . The external API keys table only exists on `develop`.

sunto (approved):

PR #514, opened 2024-06-19
[bphh-1682] hotfix to regenerate missing pdfs on user view and push websocket update

taw27:
## Description
Right now there is no error handling to retry failed generation of pdf files.  This is due a limitation of the architecture as we are using a node process to generate the files

To mitigate the problem, this PR implements a hotfix to regenerate any missing pdfs, when a user opens the submiteted permit application

- [ ] 🍕 Feature
- [x ] 🔥 Hotfix
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1682

## Steps to QA
- Difficult to manually test. Posting a video, when a missing pdf was forcefully produced via code

https://github.com/bcgov/HOUS-permit-portal/assets/32022335/89dfba68-a006-46cd-a509-6e60c4f24445

### DIFF
--- a/app/frontend/services/api/index.ts
+++ b/app/frontend/services/api/index.ts
@@ -229,7 +229,9 @@ export class Api {
   }
 
   async submitPermitApplication(id, params) {
-    return this.client.post(`/permit_applications/${id}/submit`)
+    return this.client.post<ApiResponse<IPermitApplication>>(`/permit_applications/${id}/submit`, {
+      permitApplication: params,
+    })
   }
 
   async fetchRequirementTemplates(params?: TSearchParams<ERequirementTemplateSortFields>) {
--- a/app/frontend/services/api/index.ts
+++ b/app/frontend/services/api/index.ts
@@ -224,10 +224,12 @@ export class Api {
     })
   }
 
+  async generatePermitApplicationMissingPdfs(id: string) {
+    return this.client.post<never>(`/permit_applications/${id}/generate_missing_pdfs`)
+  }
+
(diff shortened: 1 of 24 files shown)


app/frontend/services/api/index.ts
```diff
@@ -224,10 +224,12 @@ export class Api {
     })
   }
 
+  async generatePermitApplicationMissingPdfs(id: string) {
+    return this.client.post<never>(`/permit_applications/${id}/generate_missing_pdfs`)
+  }
+
   async submitPermitApplication(id, params) {
```
bkmaibach, line 231: remove params if no longer needed?

app/policies/permit_application_policy.rb
```diff
@@ -32,6 +32,10 @@ def submit?
     update?
   end
 
+  def generate_missing_pdfs?
+    true
```
bkmaibach, line 36: shouldnt there be some check that the user has the right to see this PDF

taw27: @bkmaibach you are right. My initial assumption was all authenticated user should be able to generate this permit application once submitted. But there should be a check that submitters own the permit application and reviewers belong to the same jurisdiction. Good catch

app/frontend/services/api/index.ts
taw27: was mistakenly removed from `submitPermitApplication` so putting it back

bkmaibach (approved):

PR #534, opened 2024-06-21
Fix issue accepting regional RM invite

LaurelOlson:
## Description

Fixes an issue where a logged in reviewer/review manger accepts a regional RM invitation sent to a different email than the one they are logged in with. Previously, this was upgrading the user to a regional RM but since the regional RM was not invited with the proper jurisdictions, the user only got assigned to their existing jurisdiction.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

[BPHH-1736](https://hous-bssb.atlassian.net/browse/BPHH-1736)

## Steps to QA

1. login as reviewer or review manager
2. login as admin in separate session. Invite a regional RM with an email that is NOT currently assigned to a review staff.
3. Accept the invite in the browser where the reviewer/review manager is logged in. After accepting the invite, the user should be come a regional RM and should have both the original jurisdiction and the jurisdiction the regional RM was invited to in the jurisdiction select.

## UI Accessibility Checklist

- [ ] Checked
      for accessibility errors?

## General Checklist

n/a

### DIFF
--- a/app/services/jurisdiction/user_inviter.rb
+++ b/app/services/jurisdiction/user_inviter.rb
@@ -37,8 +37,8 @@ class Jurisdiction::UserInviter
           user.invite!(inviter)
           self.results[:reinvited] << user
         elsif inviter.invitable_roles.include?(user_params[:role].to_s)
-          jurisdiction_id = user_params[:jurisdiction_id] || inviter.jurisdiction&.id
-          user = User.new(user_params.merge({ discarded_at: nil, jurisdiction_id: }))
+          jurisdiction_id = user_params.delete(:jurisdiction_id) || inviter.jurisdictions.pluck(:id)
+          user = User.new(user_params.merge({ discarded_at: nil, jurisdiction_ids: [jurisdiction_id].flatten }))
           user.skip_confirmation_notification!
           user.save!
           user.invite!(inviter)
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_06_14_002722) do
+ActiveRecord::Schema[7.1].define(version: 2024_06_21_234527) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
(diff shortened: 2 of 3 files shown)


app/services/jurisdiction/user_inviter.rb
```diff
@@ -37,8 +37,8 @@ def invite_users
           user.invite!(inviter)
           self.results[:reinvited] << user
         elsif inviter.invitable_roles.include?(user_params[:role].to_s)
-          jurisdiction_id = user_params[:jurisdiction_id] || inviter.jurisdiction&.id
-          user = User.new(user_params.merge({ discarded_at: nil, jurisdiction_id: }))
+          jurisdiction_id = user_params.delete(:jurisdiction_id) || inviter.jurisdictions.pluck(:id)
+          user = User.new(user_params.merge({ discarded_at: nil, jurisdiction_ids: [jurisdiction_id].flatten }))
```
taw27, line 41: wouldn't `inviter.jurisdictions.pluck(:id)` be an array of ids? Will this work if this is the case?

db/schema.rb
```diff
@@ -64,12 +64,6 @@
             name: "index_contacts_on_contactable"
   end
 
-  create_table "data_migrations",
-               primary_key: "version",
-               id: :string,
-               force: :cascade do |t|
-  end
-
```
taw27, line 72: This gets auto generated for the `rails data:migrate`, I am assuming it's not there if you haven't run local data migrations.  Please run rails data:migrate to put this back

taw27 (changes requested):

db/schema.rb
LaurelOlson: @taw27 weird - I did a DB dump to make sure it was in line and this got removed.. What version of PG are you on?

LaurelOlson: nvm I misread that, will do

taw27: @LaurelOlson data migrations are separate from db:migrations and don't get run automatically (intentionally). If you run rails data migrate or rails db:migrate:with_data this table will get generated along with the data migrations (which isn't relevant for clean db)

app/services/jurisdiction/user_inviter.rb
LaurelOlson: that's what the `.flatten` is for

taw27: @LaurelOlson 
Makes sense, didn't notice the flatten

db/schema.rb
LaurelOlson: OK fixed, thanks!

taw27 (approved):

PR #538, opened 2024-06-24
[bphh-1746] Refactor integration mapping creation callbacks into background jobs

taw27:
## Description
There is a callback logic to create integration mappings for a jurisdiction, when a template enables it’s API. This callback logic is very expensive and the time taken increases significantly as the template versions in the system grows.

 The immediate quick fix is to background the call backs into a background job, and this PR implements it. A future refactor will take place to further rearchitect this to improve performance for the callbacks [BPHH-1747](https://hous-bssb.atlassian.net/browse/BPHH-1747).

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

 [BPHH-1746](https://hous-bssb.atlassian.net/browse/BPHH-1746)

### DIFF
--- a/app/models/jurisdiction.rb
+++ b/app/models/jurisdiction.rb
@@ -28,7 +28,7 @@ class Jurisdiction < ApplicationRecord
   before_validation :set_type_based_on_locality
   before_save :sanitize_html_fields
 
-  after_save :create_integration_mappings, if: :saved_change_to_external_api_enabled?
+  after_save :create_integration_mappings_async, if: :saved_change_to_external_api_enabled?
 
   accepts_nested_attributes_for :contacts
   accepts_nested_attributes_for :permit_type_submission_contacts,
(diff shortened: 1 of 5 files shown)


app/models/jurisdiction.rb
```diff
@@ -162,23 +162,32 @@ def zero_carbon_step_required(activity = nil, permit_type = nil)
...
     end
   end
 
   private
 
+  def create_integration_mappings_async
+    return unless external_api_enabled?
+
+    if Rails.env.test?
+      ModelCallbackJob.new.perform(self.class.name, id, "create_integration_mappings")
```
bkmaibach, line 185: are you certain that two ModelCallbackJobs with the same class, id, and callback name will never need to be run at the same time?

bkmaibach (approved):

app/models/jurisdiction.rb
taw27: @bkmaibach thats what the lock is for. It has unique lock based on the params, which in this case would be the model id, model name and callback name. I've configured it so that it rejects any further jobs with same parameters. There is a couple of options for conflict resolution but I thought this made the most sense as we shouldn't need to run the same job with the same params. Please let me know if another default would be preferable

PR #545, opened 2024-06-25
Fix issue showing permits on submission inbox

LaurelOlson:
## Description

Fix duplicate search calls on submission inbox screen that was leading to inconsistent results when displaying permit applications for different jurisdictions as a regional RM.

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

[BPHH-1733](https://hous-bssb.atlassian.net/browse/BPHH-1733)

## Steps to QA

As a regional RM, go to the submission inbox. Verify there is only one API call made to the jurisdiction permit applications search endpoint. Change the jurisdiction and verify this also makes a single API call to fetch the permit applications for the new jurisdiction.

## UI Accessibility Checklist

- [ ] Checked
      for accessibility errors?

## General Checklist

n/a

### DIFF
--- a/app/frontend/hooks/use-search.ts
+++ b/app/frontend/hooks/use-search.ts
@@ -51,5 +51,5 @@ export const useSearch = (searchModel: ISearch, dependencyArray: any[] = []) => 
       page: searchModel.currentPage,
       countPerPage: searchModel.countPerPage,
     })
-  }, [...dependencyArray, location.search])
+  }, [...dependencyArray])
 }


app/frontend/hooks/use-search.ts
```diff
@@ -51,5 +51,5 @@ export const useSearch = (searchModel: ISearch, dependencyArray: any[] = []) =>
       page: searchModel.currentPage,
       countPerPage: searchModel.countPerPage,
     })
-  }, [...dependencyArray, location.search])
+  }, [...dependencyArray])
```
LaurelOlson, line 54: @bkmaibach any specific reasons why `location.search` was a dependency here? I can't see any reason why the location search would change without a page refresh (which would trigger the useEffect anyway) unless this was added to fix something in particular?

taw27 (approved):

PR #564, opened 2024-07-09
add aasm transitions and update email templates, notification data, substatus functionality etc

bkmaibach:
## Description

This PR adds aasm for state transitions on the permit application status.
Note that the actual status is kept at a "low resolution" (ie draft, submitted, approved only) because the minutiae about if a submitted application was viewed or if a draft application was sent back for revisions is not relevant to the bulk of the application where status is actually concerned. Therefor, the state transitions have been set up to manage timestamps, from which the "substatus" can be computed efficiently. While the "status" is kept at a high level, the "substatus" indicates what the status tag and emails should show. This separates important application backend logic from user details.

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1191

## Steps to QA

See the steps outlined in the card.
There should be notifications and emails for submission, revision request, and re-submission. This PR does not deal with the submitter's interaction with the revisions request, but only the state transitions, notifications, and status tag state.


bkmaibach:
https://github.com/bcgov/HOUS-permit-portal/pull/576

PR #576, opened 2024-07-16
Feature/bphh 1819/submission version

bkmaibach:
## Description

revised PR for new revision flow.

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://www.figma.com/design/OarUownTDVGJcI8ZqYdk0X/BC-Building-Permit-Hub?node-id=9934-87204&t=0MGp92zGUnerYLvI-0
https://hous-bssb.atlassian.net/browse/BPHH-1456
https://hous-bssb.atlassian.net/browse/BPHH-1455
https://hous-bssb.atlassian.net/browse/BPHH-1458
https://hous-bssb.atlassian.net/browse/BPHH-1191
https://hous-bssb.atlassian.net/browse/BPHH-1461
https://hous-bssb.atlassian.net/browse/BPHH-1822
https://hous-bssb.atlassian.net/browse/BPHH-1460

## Steps to QA

see cards and Figma


app/frontend/components/domains/permit-application/review-permit-application-screen.tsx
```diff
@@ -69,78 +85,131 @@ export const ReviewPermitApplicationScreen = observer(() => {
       onReferenceNumberChange(referenceNumberSnapshot)
     }
   })
-
+  const permitHeaderHeight = document.getElementById("permitHeader")?.offsetHeight
```
taw27, line 88: default to using `refs` when possible instead of DOM api in React. In this case the `permitHeader` is directly available below, so use a `ref` to access the DOM node https://react.dev/learn/manipulating-the-dom-with-refs#best-practices-for-dom-manipulation-with-refs

app/frontend/components/domains/permit-application/review-permit-application-screen.tsx
```diff
@@ -69,78 +85,131 @@ export const ReviewPermitApplicationScreen = observer(() => {
...
+            bg="theme.yellow"
+            justify="flex-start"
+            align="center"
+            gap={4}
+            top={permitHeaderHeight}
           >
-            {t("ui.backToInbox")}
-          </Button>
-        </Stack>
+            <Flex width="var(--app-sidebar-width)" align="center" gap={2}>
```
taw27, line 161: can add this variable directly `app/frontend/styles/theme/foundations/sizes.ts`

app/frontend/components/domains/permit-application/revision-sidebar.tsx
```diff
@@ -0,0 +1,248 @@
...
+      setTopHeight(permitHeaderHeight)
+    }, [isMounted, window.innerWidth, window.innerHeight])
+
+    const { isOpen, onOpen, onClose } = useDisclosure()
+
+    const onFinalizeRevisions = () => {
+      ;(async () => {
+        const ok = await permitApplication.finalizeRevisionRequests()
+        if (ok) navigate(`/jurisdictions/${permitApplication.jurisdiction.slug}/submission-inbox`)
+      })()
```
taw27, line 78: what's the point of the IIFE? Just make the top level function async do the same thing in the function body

taw27 (changes requested):

app/frontend/components/shared/permit-applications/permit-application-status-tabs.tsx
```diff
@@ -14,41 +14,48 @@ export const PermitApplicationStatusTabs = observer(function ToggleArchivedButto
...
-    return index === -1 ? 0 : index
+  const draftFilters = [EPermitApplicationStatus.newDraft, EPermitApplicationStatus.revisionsRequested]
+  const submittedFilters = [EPermitApplicationStatus.newlySubmitted, EPermitApplicationStatus.resubmitted]
+
+  const statusStringToIndex = (statusString: string): number => {
+    const map = {
+      [draftFilters.join(",")]: 0,
+      [submittedFilters.join(",")]: 1,
+    }
+    return map[statusString]
```
taw27, line 28: what if the status string order is not as expected. Since it's a queryParam, anyone can edit those in the url bar

taw27 (changes requested):

app/frontend/utils/formio-submission-traversal.ts
```diff
@@ -0,0 +1,33 @@
...
+    Object.keys(submissionData.data).forEach((sectionKey) => {
+      const section = submissionData.data[sectionKey]
+      if (!R.isEmpty(data)) return
+
+      const filteredObject = Object.keys(section)
+        .filter((key) => key.startsWith(requirementKey))
+        .reduce((filteredObj, key) => {
+          filteredObj[key] = section[key]
+          return filteredObj
+        }, {})
```
taw27, line 16: not necessary, but can use `https://ramdajs.com/docs/#pickBy` to make this cleaner. Alternatively the filter is not required at all, the same logic can be added directly to the `reduce`.

```
const filteredObject = Object.keys(section)
        .reduce((filteredObj, key) => {
          if(!key.startsWith(requirementKey)) return filteredObj

          filteredObj[key] = section[key]
          return filteredObj
        }, {})
```

app/frontend/utils/formio-submission-traversal.ts
```diff
@@ -0,0 +1,33 @@
...
+    })
+    return { data }
+  } else {
+    let foundSubmission = null
+
+    Object.keys(submissionData.data).forEach((sectionKey) => {
+      if (submissionData.data[sectionKey][requirementKey]) {
+        foundSubmission = submissionData.data[sectionKey][requirementKey]
+      }
+    })
```
taw27, line 30: same as previous comment

app/frontend/utils/formio-helpers.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  return {
+    id: requirementJson.id,
+    key: `single-${requirementJson.key}`,
+    components: [requirementJson],
+  }
+}
+
+export const isFieldSetKey = (key: string) => {
+  return key.split("|").length === 4
```
taw27, line 15: Not sure of the exact use case, but this might be a bit flaky, as if anyone introduces new  key with the same split length it will break. If there is a more deterministic way, I would prefer using that, otherwise this is is fine

app/frontend/stores/permit-application-store.ts
```diff
@@ -138,11 +151,11 @@ export const PermitApplicationStoreModel = types
     addPermitApplication(permitapplication: IPermitApplication) {
       self.permitApplicationMap.put(permitapplication)
     },
-    setStatusFilter(status: TFilterableStatus) {
-      if (!status) return
-
-      setQueryParam("status", status)
-      self.statusFilter = status
+    setStatusFilter(statuses: TFilterableStatus[]) {
+      if (!statuses) return
```
taw27, line 155: update type to be statuses: TFilterableStatus[] | undefined to keep it consistent with this conditional

app/frontend/i18n/i18n.ts
```diff
@@ -1625,9 +1681,9 @@ const options = {
   lng: "en", // default language
   fallbackLng: "en",
   interpolation: { escapeValue: false },
-  parseMissingKeyHandler: (key, defaultValue) => {
-    return fallbackTranslations.en.fallback
-  },
+  // parseMissingKeyHandler: (key, defaultValue) => {
+  //   return fallbackTranslations.en.fallback
+  // },
```
taw27, line 1686: remove or add a TODO if this needs to be hadled later

app/frontend/components/shared/revisions/revision-modal.tsx
```diff
@@ -0,0 +1,211 @@
...
+      } else {
+        // Item does not exist, append it
+        append(newItem)
+      }
+
+      onSave().then(() => {
+        elements?.[0]?.classList?.add("revision-requested")
+        handleClose()
+      })
+    }
```
taw27, line 100: instead of having nested conditionals. It is better to early return when possible for better readability

taw27 (changes requested):

app/frontend/components/domains/permit-application/edit-permit-application-screen.tsx
```diff
@@ -330,7 +331,7 @@ export const EditPermitApplicationScreen = observer(({}: IEditPermitApplicationS
               gap={4}
               top={permitHeaderHeight}
             >
-              <Flex width="var(--app-sidebar-width)" align="center" gap={2}>
+              <Flex width={sizes.sidebar.width} align="center" gap={2}>
```
taw27, line 334: @bkmaibach if you have added this to the sizes, you can directly use `width={"sidebar.width"}`

app/models/concerns/permit_application_status.rb
```diff
@@ -0,0 +1,57 @@
+module PermitApplicationStatus
+  extend ActiveSupport::Concern
+  included do
+    include AASM
+    enum status: { new_draft: 0, newly_submitted: 1, revisions_requested: 3, resubmitted: 4 }, _default: 0
```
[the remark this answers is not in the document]
bkmaibach: yeah that's what I was trying to push forward, but the mocks need to show if a permit application was revision requested, or resubmitted.
Originally I was determining this using timestamps, but Sunny and Laurel did not let me put this through and said I needed to refactor this with the new states, leading to these somewhat redundant states.
The reason new_draft and newly_submitted is to distinguish it from the other draft and submitted states respectively. This is so that these are not confused with the filters on the front end which do not correspond to actual statuses, but status groups.

app/frontend/components/domains/permit-application/review-permit-application-screen.tsx
```diff
@@ -158,7 +164,7 @@ export const ReviewPermitApplicationScreen = observer(() => {
             gap={4}
             top={permitHeaderHeight}
           >
-            <Flex width="var(--app-sidebar-width)" align="center" gap={2}>
+            <Flex width={sizes.sidebar.width} align="center" gap={2}>
```
taw27, line 167: once added to sizes, you can just use `width={"sidebar.width"}`

taw27 (changes requested):

app/models/concerns/permit_application_status.rb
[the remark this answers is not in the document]
taw27: ok. Personally like the timestamp approach if the only need is to show that  data has been resubmitted, but I don't have enough context so assuming it has been discussed in detail already.

app/frontend/utils/formio-helpers.ts
bkmaibach: the form is just like this, I don't think there is a better way

app/frontend/components/domains/permit-application/edit-permit-application-screen.tsx
```diff
@@ -184,117 +199,165 @@ export const EditPermitApplicationScreen = observer(({}: IEditPermitApplicationS
...
+    }
+  }, [currentPermitApplication?.isRevisionsRequested])
+
   if (error) return <ErrorScreen error={error} />
   if (!currentPermitApplication?.isFullyLoaded) return <LoadingScreen />
 
-  const scrollToBottom = () => {
-    handleScrollToBottom()
-  }
+  const permitHeaderHeight = document.getElementById("permitHeader")?.offsetHeight
```
taw27, line 216: useRef? Unless not used for other reasons

db/migrate/20240621234528_create_revision_requests.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    def data
+      PermitApplication.submitted.each do |pa|
+        pa.submission_versions.create(
+          form_json: pa.form_json,
+          submission_data: pa.submission_data,
+          created_at: pa.submitted_at,
+          viewed_at: pa.viewed_at,
+        )
+      end
```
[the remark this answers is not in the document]
taw27: change this to `PermitApplication.where(status: 1)`. I know you added ` scope :submitted, -> { joins(:submission_versions).distinct }` to the permit_application model, but the above needs the original status column, as it's copying over the values from the original model to this new model

db/data/20240621234529_create_submission_versions.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class CreateSubmissionVersions < ActiveRecord::Migration[7.1]
+  def up
+    PermitApplication.submitted.each do |pa|
```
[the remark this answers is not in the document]
taw27: change this to PermitApplication.where(status: 1). I know you added  scope :submitted, -> { joins(:submission_versions).distinct } to the permit_application model, but the above needs the original status column, as it's copying over the values from the original model to this new model

taw27 (changes requested):

taw27 (approved):

PR #578, opened 2024-07-16
Only require step code if requirements exist

LaurelOlson:
## Description

Refactor form json into a separate service and add in functionality to transform the step code so it is required or not depending on the presence of step requirements.

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

[BPHH-1489](https://hous-bssb.atlassian.net/browse/BPHH-1489)

## Steps to QA

Login as a review manager and update step requirements for a permit type so both are "not required". This should remove the validation on the step code field in the permit application form. Setting either energy or zero carbon requirements back to a value should make the step code required again. 

## UI Accessibility Checklist

- [ ] Checked
      for accessibility errors?

## General Checklist

n/a

### DIFF
--- a/app/models/permit_application.rb
+++ b/app/models/permit_application.rb
@@ -48,18 +48,8 @@ class PermitApplication < ApplicationRecord
   scope :unviewed, -> { where(status: :submitted, viewed_at: nil).order(submitted_at: :asc) }
 
   def form_json
-    processed_json = template_version.form_json.deep_dup
-    block_ids_to_remove = get_requirement_block_ids_to_remove
-
-    return processed_json if block_ids_to_remove.empty?
-
-    processed_json["components"].each do |component|
-      next unless component["components"].present?
-
-      component["components"].delete_if { |sub_component| block_ids_to_remove.include?(sub_component["id"]) }
-    end
-
-    remove_empty_sections_from_form_json!(processed_json)
+    result = PermitApplication::FormJsonService.new(permit_application: self).call
+    result.form_json
   end
 
   def force_update_published_template_version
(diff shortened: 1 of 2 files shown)


app/models/permit_application.rb
```diff
@@ -217,6 +207,14 @@ def missing_pdfs
     missing_pdfs
   end
 
+  def energy_step_code_required?
+    jurisdiction
+      .permit_type_required_steps
+      .where(permit_type_id:)
+      .customizations
+      .all? { |rs| !rs.energy_step_required && !rs.zero_carbon_step_required }
```
taw27, line 215: so all of energy_step_required and zero_carbon_step_required have to be false for energy step_code to be required?

taw27 (approved):

app/models/permit_application.rb
LaurelOlson: The real world expectation is that if it's not required, there will only be one customization and both values will be nil. It wouldn't make sense if there are multiple required steps where one of them has both set to not required but if that is the case, we would treat it as required as there is a set of requirements.

LaurelOlson: OK actually yes this logic is flawed, will update

LaurelOlson: @taw27 OK fixed, should be easier to make sense of now and works correctly

taw27: @LaurelOlson what happens if only one  of `energy_step_required ` or `zero_carbon_step_required ` is nil? This  would be energy_step_code_required? is false?

taw27 (approved):

app/models/permit_application.rb
taw27: @LaurelOlson approved after latest change

PR #579, opened 2024-07-16
Feature/bphh 1540/configure reasons

bkmaibach:
## Description

Allow the super admin to create custom reasons for revision under configuration management

- only 20 reasons are allowed (10 may be too restrictive)
- revisions can be deleted (archived) but are still referenced by existing revision requests
- only 200 reason codes can exist in total (archived and un-archived) as all of these are loaded in the site configuration
- revision requests reference reasons by reason_code, this is the identity of the reason
- reasons re-created under a previously archived reason code is equivalent to un-archiving and updating the reason
- reasons description can be edited and the new text will appear wherever that reason code is referenced

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1540

## Steps to QA

See card. You might also want to check that reason text changes when looking at a revision requested permit application.


app/models/site_configuration.rb
```diff
@@ -3,14 +3,66 @@ class SiteConfiguration < ApplicationRecord
...
 
   def self.instance
     first_or_create
   end
 
+  def revision_reasons_attributes=(attributes)
+    attributes.each do |attribute, _|
+      if attribute["_destroy"] == true
+        # Mark the record for destruction
+        self.revision_reasons.find(attribute["id"])&.discard if attribute["id"].present?
```
taw27, line 22: the `find` method will throw an error if the `revision_reasons` is not found. If this is not intentional change the method to `find_by`

app/models/site_configuration.rb
```diff
@@ -3,14 +3,66 @@ class SiteConfiguration < ApplicationRecord
...
+
   HELP_LINK_KEYS = %w[get_started_link_item best_practices_link_item dictionary_link_item user_guide_link_item]
 
   def self.instance
     first_or_create
   end
 
+  def revision_reasons_attributes=(attributes)
+    attributes.each do |attribute, _|
+      if attribute["_destroy"] == true
```
taw27, line 20: The `_destroy`, method is used by rails active record to mark nested attributes for destruction. If the goal is to archive, use a different property name to keep with convention. Better yet, have a`should_discard` or similar attribute accessor in the `revision_reason` model, and pass this flag from the front-end. In the `revision_reason` model have a `before_save` callback to set the `discarded_at` if this attribute is `true`

app/models/site_configuration.rb
```diff
@@ -3,14 +3,66 @@ class SiteConfiguration < ApplicationRecord
...
+        # Mark the record for destruction
+        self.revision_reasons.find(attribute["id"])&.discard if attribute["id"].present?
+      else
+        if attribute["id"].present?
+          # Update the existing record
+          existing_record = self.revision_reasons.with_discarded.find(attribute["id"])
```
taw27, line 26: the find will cause an an exception to be raised if not found. Similar to the `_destroy` comment, this would be better done using some kind of attribute accessor instead of doing this manipulation here as the front-end should know if something has been archived or not

app/models/site_configuration.rb
```diff
@@ -3,14 +3,66 @@ class SiteConfiguration < ApplicationRecord
...
+            # Undiscard and update the record
+            discarded_reason.undiscard
+            discarded_reason.update(attribute.except("id"))
+          else
+            # Proceed with the normal behavior
+            self.revision_reasons.build(attribute)
+          end
+        end
+      end
+    end
```
taw27, line 43: This whole method would be better to extract away in a service.

db/data/20240716232716_create_revision_reasons.rb
```diff
@@ -0,0 +1,22 @@
...
+    site_configuration = SiteConfiguration.instance
+
+    site_configuration.revision_reasons.create(
+      [
+        { reason_code: "zoning_non_compliance", description: "Zoning non-compliance" },
+        { reason_code: "inaccurate_documentation", description: "Inaccurate documentation" },
+        { reason_code: "failure_to_meet_building_code", description: "Failure to meet building code requirements" },
+        { reason_code: "other", description: "Other" },
+      ],
+    )
```
taw27, line 14: This should probably be a part of seeds.

taw27 (changes requested):

app/models/site_configuration.rb
bkmaibach: Isnt the idea to override the existing destruction behaviour? If I go RevisionReason.first.destroy, that would typically destroy the record, but the replaced behaviour is already abstracted away for me. Why would I need to make this distinction? Wouldnt this create ambiguisty?

app/models/site_configuration.rb
bkmaibach: I'm not sure what you mean

db/data/20240716232716_create_revision_reasons.rb
bkmaibach: Are seeds populated on production? This is for production.

taw27: @bkmaibach double check with Sunny but I think this should go into production seeds

taw27 (approved):

PR #585, opened 2024-07-18
add discard field to reasons and refactor attribute assigner

bkmaibach:
Quick update to use a mark for discard structure for revision reasons as per Taw's suggestion. compact and simplify the attribute assign logic.


app/models/site_configuration.rb
```diff
@@ -17,29 +17,21 @@ def self.instance
...
-            # Proceed with the normal behavior
-            self.revision_reasons.build(attribute)
-          end
+        if existing_reason.present?
+          # Undiscard and update the record
+          existing_reason.update(attribute.except("id").merge({ discarded_at: nil }))
+        else
+          # Proceed with the normal behavior
+          self.revision_reasons.build(attribute)
         end
```
taw27, line 35: replace this with and see if it works. If not original way is fine but this way the update is done by rails active record
```
 def revision_reasons_attributes=(attributes)
    attributes.each do |attribute, _|
      next unless attribute["id"].blank? && attribute["reason_code"].present?

      existing_reason = self.revision_reasons.with_discarded.find_by(reason_code: attribute["reason_code"])

      next unless existing_reason.present?

      # Undiscard and update the record
      attribute["id"] = existing_reason.id
      attribute["discarded_at"] = nil
    end

    super(attributes)
  end
```

app/models/revision_reason.rb
```diff
@@ -9,9 +9,18 @@ class RevisionReason < ApplicationRecord
...
+  before_save :discard_if_marked
+
   private
 
   def normalize_reason_code
     self.reason_code = reason_code.parameterize(separator: "_") if reason_code.present?
   end
+
+  def discard_if_marked
+    return if discarded_at.present?
```
taw27, line 22: check if there was a change to discard_at if if discarded was initially nil. Other wise this might update the discarded_at to a new value when it wasn't intended to

taw27 (changes requested):

app/models/site_configuration.rb
taw27: remember when using nested attributes, have to call save on the parent model

app/models/revision_reason.rb
bkmaibach: good catch

bkmaibach: actually, I believe this is already prevented by the return if discarded_at.present? guard clause
this function will never update discarded_at if it is present.

taw27: okk I misread. Should be fine

taw27 (approved):

taw27 (approved):

PR #600, opened 2024-07-26
add special designations first nations templates and blocks, and copying

bkmaibach:
## Description

Full first nation feature with everything for copying from exisitng template and importing customizations.
shows new column in requirement template and block.
Shows first nation tag on block
etc.

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1780
https://hous-bssb.atlassian.net/browse/BPHH-1782
https://hous-bssb.atlassian.net/browse/BPHH-1783
https://hous-bssb.atlassian.net/browse/BPHH-1785
https://hous-bssb.atlassian.net/browse/BPHH-1658
https://hous-bssb.atlassian.net/browse/BPHH-1781
https://hous-bssb.atlassian.net/browse/BPHH-1784

## Steps to QA

See cards and mocks


app/frontend/models/template-version.ts
```diff
@@ -94,6 +100,40 @@ export const TemplateVersionModel = types
 
       return response
     }),
+    copyJurisdictionTemplateVersionElectives: flow(function* (jurisdictionId: string) {
+      const response = yield* toGenerator(
+        self.environment.api.copyJurisdictionTemplateVersionCustomization(self.id, jurisdictionId, true, false, true)
+      )
+      if (!response.ok) {
+        return response
```
taw27, line 108: just verifying if the whole response object was intended to be returned or just true/ false

app/models/requirement_template.rb
```diff
@@ -33,12 +33,16 @@ class RequirementTemplate < ApplicationRecord
   # This is a workaround needed to validate step code related errors
   attr_accessor :requirement_template_sections_attributes_copy
 
-  validate :unique_permit_and_activity_for_undiscarded, on: :create
+  validate :unique_classification_for_undiscarded, on: :create
   validate :validate_uniqueness_of_blocks
   validate :validate_step_code_related_dependencies
 
+  def requirement_template_label
+    requirement_template.label
```
taw27, line 41: if this is the requirement_template mode, what is this `requirement_template` referring to here. Should this be `self` instead?

app/services/requirement_template_copy_service.rb
```diff
@@ -0,0 +1,36 @@
...
+      # Save the new template first to get its ID
+      new_template.save!
+      # Clone the sections and associate them with the new template
+      requirement_template.requirement_template_sections.each do |section|
+        new_section = new_template.requirement_template_sections.build(name: section.name, position: section.position)
+
+        section.save!
+
+        # Associate existing requirement blocks with the new section
+        section.requirement_blocks.each { |block| new_section.requirement_blocks << block }
```
taw27, line 28: no need to call `section.save` after this?

taw27 (changes requested):

app/models/requirement_template.rb
bkmaibach: I think this was added by mistake, it doesnt seem to be getting used anywhere

app/services/requirement_template_copy_service.rb
bkmaibach: Good point!

taw27 (approved):

PR #635, opened 2024-08-19
[bphh-1987] Misc collaboration  fixes

taw27:
The original issue mentioned in the card  seems to be just a reindex issue. I'm not sure how it got there, as it's reindexed properly locally, but I just reindexed manually on the deployed environment so there is no code change for that.

While debugging I found some other tangentially related issues, which I is included in this PR

- Making the assignment popover lazy so that the assignment search only occurs when the popover is open reducing the number of round-trip calls

- Include permit_collaboration eagerly for permit_application search

- Callbacks to create/ destroy jurisdiction collaborators when a user is undiscarded/ discarded and when there is a change to their jurisdiction memberships

- [ ] 🍕 Feature
- [x ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1987

## Steps to QA

The other fixes happen in the background, so not a lot to test.

- When a new user is added or unarchived to a jurisdiction. they should be available in the assignment popover above

- When a user is archived or if they are removed from a jurisdiction, they should no longer be available in the assignment popover above

### DIFF
--- a/app/models/jurisdiction_membership.rb
+++ b/app/models/jurisdiction_membership.rb
@@ -3,8 +3,28 @@ class JurisdictionMembership < ApplicationRecord
   belongs_to :user
 
   after_commit :reindex_jurisdiction, on: %i[create update destroy]
+  after_commit :create_jurisdiction_collaborator, on: %i[create update]
+  after_commit :destroy_jurisdiction_collaborator, on: %i[update]
 
   private
+
+  def create_jurisdiction_collaborator
+    return unless user.review_staff? && user.kept?
+
+    existing_collaborator = jurisdiction.collaborators.find_by(user_id: user.id)
+
+    return if existing_collaborator.present?
+
+    jurisdiction.collaborators.create(user: user)
+  end
+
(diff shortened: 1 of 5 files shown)


app/models/jurisdiction_membership.rb
```diff
@@ -3,9 +3,29 @@ class JurisdictionMembership < ApplicationRecord
   belongs_to :user
 
   after_commit :reindex_jurisdiction, on: %i[create update destroy]
+  after_commit :create_jurisdiction_collaborator, on: %i[create update]
+  after_commit :destroy_jurisdiction_collaborator, on: %i[update]
```
bkmaibach, line 7: both create and destroy jurisdiction collaborator on update? Is this right?

taw27: @bkmaibach good catch. Mistake from copying over previous line

bkmaibach (approved):

PR #646, opened 2024-08-22
[bphh-2006] Fix bug with force publish still calling synchronus publi…

taw27:
## Description
Fixes bug with force publishes still calling synchronous publish method. Also, fixes force publish returned template versions

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-2006

## Steps to QA
- Go to any template and force publish, note the template might say "scheduled" initially, but websocket will update it's state

### DIFF
--- a/app/services/template_versioning_service.rb
+++ b/app/services/template_versioning_service.rb
@@ -92,8 +92,6 @@ class TemplateVersioningService
 
     ModelCallbackJob.perform_async(template_version.class.name, template_version.id, "force_publish_now!")
 
-    template_version = publish_version!(template_version, true)
-
     template_version
   end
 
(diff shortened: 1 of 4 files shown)


app/services/template_versioning_service.rb
```diff
@@ -92,8 +92,6 @@ def self.force_publish_now!(requirement_template)
 
     ModelCallbackJob.perform_async(template_version.class.name, template_version.id, "force_publish_now!")
 
-    template_version = publish_version!(template_version, true)
```
bkmaibach, line 95: @taw27 Hoping you could help me understand why this line was removed

taw27: @bkmaibach see previous line. This method is backgrounded already, but I forgot to remove the synchronous line before ` template_version = publish_version!(template_version, true)` when I did the background job feature

bkmaibach (approved):

PR #648, opened 2024-08-22
[BPHH-1982] CSRF Token Fix

derekyau:
## Description

…in CSRF protection so that we would be safe from those attacks. This change makes the backend set a CSRF-TOKEN cookie which is then read on the requestTransform in the frontend and set so that we can maintain security

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

Implements https://hous-bssb.atlassian.net/jira/software/projects/BPHH/boards/1?issueParent=10320&selectedIssue=BPHH-1982

## Steps to QA
As long as you can login as use the app (submit various forms, etc.) as usual there should be no impact on user experience. As a developer, inspect the headers to see that CSRF token is indeed passed.

### DIFF
--- a/app/controllers/api/application_controller.rb
+++ b/app/controllers/api/application_controller.rb
@@ -1,6 +1,10 @@
 class Api::ApplicationController < ActionController::API
+  include ActionController::Cookies
+  include ActionController::RequestForgeryProtection
   include BaseControllerMethods
   include Pundit::Authorization
+
+  protect_from_forgery with: :exception
 
   before_action :authenticate_user!
   before_action :check_for_archived_user
@@ -11,6 +15,7 @@ class Api::ApplicationController < ActionController::API
 
   after_action :verify_authorized, except: %i[index], unless: :skip_pundit?
   after_action :verify_policy_scoped, only: %i[index], unless: :skip_pundit?
+  after_action :set_csrf_cookie
 
   def index
     # This parent application controller throws errors from above after_actions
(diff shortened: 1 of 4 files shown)


app/controllers/api/application_controller.rb
```diff
@@ -11,17 +15,31 @@ class Api::ApplicationController < ActionController::API
...
   def index
     # This parent application controller throws errors from above after_actions
     # if the subclass does not implement this method. This is provided as a fallback.
     raise NotImplementedError, "The index method is not implemented."
   end
 
+  private
+
+  def set_csrf_cookie
+    cookies["CSRF-TOKEN"] = {
```
bkmaibach, line 29: I'm a little confused by this.
I thought the whole point of CSRF protection was to not rely on cookies?
Isn't the idea that the cookie is automatically associated and attached by the browser based on the destination of the request, not the origin of the request? Is this cookie set every time the HTTP only auth is set?

bkmaibach (approved):

bkmaibach:
https://blog.eq8.eu/article/rails-api-authentication-with-spa-csrf-tokens.html

app/controllers/api/application_controller.rb
derekyau: See our conversation here for more details: https://laterolabs.slack.com/archives/CP688FUUQ/p1724361988634719

PR #660, opened 2024-08-26
fix filters in pa strong params, and use of filter query params

bkmaibach:
## Description

Filters were not being properly applied when clicking the link in template update notifications.

- [ ] 🍕 Feature
- [x ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-2030

### DIFF
--- a/app/controllers/api/concerns/search/permit_applications.rb
+++ b/app/controllers/api/concerns/search/permit_applications.rb
@@ -17,7 +17,6 @@ module Api::Concerns::Search::PermitApplications
         ),
       includes: PermitApplication::SEARCH_INCLUDES,
     }
-
     @permit_application_search = PermitApplication.search(permit_application_query, **search_conditions)
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/api/concerns/search/permit_applications.rb
```diff
@@ -28,11 +27,7 @@ def permit_application_search_params
       :query,
       :page,
       :per_page,
-      filters: {
-        status: [],
-        template_version_id: :template_version_id,
-        requirement_template_id: :requirement_template_id,
-      },
+      filters: [:requirement_template_id, :template_version_id, { status: [] }],
```
taw27, line 30: is the brackets around status intentional

bkmaibach: yes, this is the only way I can get this to work (it's weird in rails 7.1 i think)

PR #672, opened 2024-08-28
update matching on permit applications

bkmaibach:
## Description

Update fields on permit application search to use different kinds of matching. Now follows searchkick >5.3 docs.

- [ ] 🍕 Feature
- [ x] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-2038

### DIFF
--- a/app/models/permit_application.rb
+++ b/app/models/permit_application.rb
@@ -7,8 +7,8 @@ class PermitApplication < ApplicationRecord
 
   SEARCH_INCLUDES = %i[permit_type submission_versions step_code activity jurisdiction submitter permit_collaborations]
 
-  searchkick searchable: %i[number nickname full_address permit_classifications submitter status review_delegatee_name],
-             word_start: %i[number nickname full_address permit_classifications submitter status review_delegatee_name]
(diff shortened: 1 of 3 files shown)


app/models/permit_application.rb
```diff
@@ -7,8 +7,8 @@ class PermitApplication < ApplicationRecord
 
   SEARCH_INCLUDES = %i[permit_type submission_versions step_code activity jurisdiction submitter permit_collaborations]
 
-  searchkick searchable: %i[number nickname full_address permit_classifications submitter status review_delegatee_name],
-             word_start: %i[number nickname full_address permit_classifications submitter status review_delegatee_name]
+  searchkick word_middle: %i[nickname full_address permit_classifications submitter status review_delegatee_name],
```
taw27, line 10: I looked at the latest docs. Instead of word_middle, wouldn't it be better to use `text_middle` ? As things like nickname, fill address etc will have spaces between them

taw27 (approved):

PR #675, opened 2024-08-28
dont use disabled classifications in frontend

bkmaibach:
## Description

This is to properly treat disabled permit and work types IE don't show them in any part of the UI

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-2040


app/controllers/api/permit_classifications_controller.rb
```diff
@@ -31,9 +33,9 @@ def permit_classification_options
           ].nil?
 
           # &:activities or &:permit_types
-          query.map(&classification_option_params[:type].underscore.to_sym).uniq
+          query.map(&classification_option_params[:type].underscore.to_sym).filter { |c| c.enabled? }.uniq
```
taw27, line 36: Wouldn't it be better to add them to the queries instead?

taw27 (approved):

PR #687, opened 2024-08-30
Chore/first impression improvement

bkmaibach:
Rebased P{R onto main for landing hotfix


karenkun (approved):
Yes I think re-adjusting the image worked well here :)

PR #693, opened 2024-09-18
Feature/bphh 2100/sandbox backend

bkmaibach:
## Description

initial backend work for sandboxes. Permit applications and customizations can belong to a sandbox.


app/controllers/external_api/v1/permit_applications_controller.rb
```diff
@@ -47,6 +47,10 @@ def show_integration_mapping
 
   private
 
+  def set_permit_application
+    @jurisdiction = Jurisdiction.find(current_external_api_key.jurisdiction_id)
+  end
```
LaurelOlson, line 52: this is conflicting now with the method below - not sure which one is correct but should only have one

app/models/sandbox.rb
```diff
@@ -0,0 +1,5 @@
+class Sandbox < ApplicationRecord
+  belongs_to :jurisdiction
```
LaurelOlson, line 2: add the `has_many` for `permit_applications` and `jurisdiction_template_version_customizations`

db/data/20240918210326_create_and_index_default_sandboxes.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CreateAndIndexDefaultSandboxes < ActiveRecord::Migration[7.1]
+  def up
+    Jurisdiction.find_each(&:save)
+    Jurisdiction.reindex
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
LaurelOlson, line 10: probably don't need to actually raise an error here

db/schema.rb
```diff
@@ -116,6 +116,24 @@
...
+    t.datetime "processed_at"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index %w[notifiable_type notifiable_id],
+            name: "index_integration_mapping_notifications_on_notifiable"
+    t.index ["template_version_id"],
+            name:
+              "index_integration_mapping_notifications_on_template_version_id"
+  end
+
```
LaurelOlson, line 136: might need to rebase this on `develop` to get rid of these changes (as this PR went in already I think?)

PR #703, opened 2024-09-26
Story/bphh 1845/toggle sandbox mode and return to live mode

bkmaibach:
## Description

Lots of work towards the review staff sandbox mode including:

- two default sandboxes per jurisdiction, scoped to published and scheduled template versions
- a sandbox selector instead of a switch
- admin sandbox select in permit application creation
- refactored useTemplateVersion hook for sandbox scoping
- refactored template version policy for sandbox scoping
- refactored application policy base class for user_contexts with sandbox
- Add LTSA error message
- handle frontend store and models for sandboxes
- handle per-jurisdiction sandbox options
- upgrade the beforeMergeUpdate behaviour of jurisdictions to handle sandboxes
- Specs
- Handle logic for adding correct sandbox ID onto newly created permit applications
- addRequestTransform for providing the currently selected sandbox ID as a header

- [x ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-1845


app/frontend/stores/permit-application-store.ts
```diff
@@ -184,8 +184,8 @@ export const PermitApplicationStoreModel = types
     },
   }))
   .actions((self) => ({
-    createPermitApplication: flow(function* (formData: TCreatePermitApplicationFormData) {
-      const { ok, data: response } = yield self.environment.api.createPermitApplication(formData)
+    createPermitApplication: flow(function* (formData: TCreatePermitApplicationFormData, sandboxId: string = null) {
+      const { ok, data: response } = yield self.environment.api.createPermitApplication(formData, sandboxId)
```
LaurelOlson, line 188: does this work? it looks like `createPermitApplication` takes in a single `params` argument so I would expect `sandboxId` would need to be merged with `formData` in order for this to get passed correctly?

app/frontend/stores/root-store.ts
```diff
@@ -58,8 +58,8 @@ export const RootStoreModel = types
         storage: localStorage,
       })
       yield makePersistable(self.sandboxStore, {
-        name: `${self.userStore.currentUser?.id}-SandboxStore`,
-        properties: ["sandboxId"],
+        name: `SandboxStore`,
+        properties: ["currentSandboxId", "active"],
```
LaurelOlson, line 62: what is `active` referring to?

app/models/permit_application.rb
```diff
@@ -415,6 +417,29 @@ def energy_step_code_required?
...
+    Jurisdiction.all.map do |j|
+      row =
+        PermitApplication.joins(:submitter).where(jurisdiction: j, users: { role: "submitter" }).group(:status).count
+
+      {
+        "draft" => (row["new_draft"] || 0) + (row["revisions_requested"] || 0), # nil defaults to 0
+        "submitted" => (row["newly_submitted"] || 0) + (row["resubmitted"] || 0),
+      }.merge({ name: j.qualified_name })
+    end
+  end
```
LaurelOlson, line 441: duplicate method

app/models/customization.rb
```diff
@@ -0,0 +1,3 @@
+# Convenience class alias to prevent carpal tunnel syndrome
+# see: app/models/jurisdiction_template_version_customization.rb
+Customization = JurisdictionTemplateVersionCustomization
```
LaurelOlson, line 3: lol we should maybe think about name-spacing models.. I might try [this](https://blog.bullettrain.co/rails-model-namespacing/) with step code and see how it works

app/frontend/stores/permit-application-store.ts
bkmaibach: this has been removed in a later branch. sandBox ID should always come from the header

app/models/permit_application.rb
bkmaibach: this needs to be removed altogether as there is now an advanced version of this on develop

app/models/customization.rb
bkmaibach: OK keep me posted

app/frontend/stores/permit-application-store.ts
LaurelOlson: OK - in the future, try to keep changes/fixes in the same PR if the PR hasn't been merged into dev yet - this makes review a lot easier

LaurelOlson (approved):

PR #715, opened 2024-10-09
Backend STI, policy, search, controller, validations etc for EarlyAccess

bkmaibach:
## Description

Backend work for re-using RequirementTemplate model to handle the concept of previews.
Adds new fields for blocks and requirement templates
Adds early_access to frontend search model and search concern for obtaining records considered early access.

This builds on the sandbox backend work, but disables it on the front end as this feature is delayed until after super admin early access feature is complete.


app/frontend/components/domains/super-admin/early-access/requirements-library/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+            <Text color={"text.secondary"} mt={1}>
+              {t("earlyAccessRequirementsLibrary.index.description")}
+            </Text>
+          </Box>
+          {/* <RequirementsBlockModal /> */}
+        </Flex>
+
+        {/* <RequirementBlocksTable alignItems={"flex-start"} w={"full"} /> */}
+
+        {/* <ToggleArchivedButton searchModel={requirementBlockStore} mt={3} /> */}
```
LaurelOlson, line 28: crud? or TODO?

app/models/early_access_requirement_template.rb
```diff
@@ -0,0 +1,23 @@
+class EarlyAccessRequirementTemplate < RequirementTemplate
+  # has_one :template_version, dependent: :destroy, foreign_key: :requirement_template_id
```
LaurelOlson, line 2: crud?

app/frontend/components/domains/super-admin/early-access/requirements-library/index.tsx
bkmaibach: todo - see later branches

app/models/early_access_requirement_template.rb
bkmaibach: TODO - fixed later

LaurelOlson (approved):

PR #734, opened 2024-10-21
Chore/fix lint staged

bkmaibach:
## Description

This consolidates our ruby formatting to use syntax_tree ONLY.
Prior to this we were set up to have syntax_tree formatting apply on save, and then prettier ruby formatting applied in lint-staged which led to confusing behaviour and large diffs.

This is being applied to the end of my preview branch to avoid merge conflict hell. We may want to consider applying this directly to develop ALSO.

To do this:

git cherry-pick 341383f35c21122f1b0fba481d82da57329df1dd
stree write ./app/**/*.rb

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,0 @@
-# Ignore auto-generated files
-db/schema.rb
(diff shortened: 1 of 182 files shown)


bkmaibach:
We will apply this directly to develop once everything is consolidated

PR #737, opened 2024-10-22
refactor early access filter as visibility and finish early access re…

bkmaibach:
## Description

Refactor early access search flag as a full string to allow searching for different combinations of any, live, and early_access.
Create a page for requirement blocks that uses a particular store for early access requirement blocks. 

https://hous-bssb.atlassian.net/browse/BPHH-2149

### DIFF
--- a/app/controllers/api/concerns/search/requirement_templates.rb
+++ b/app/controllers/api/concerns/search/requirement_templates.rb
@@ -5,17 +5,10 @@ module Api::Concerns::Search::RequirementTemplates
     @search =
       RequirementTemplate.search(
         query,
-        includes: %i[
-          activity
-          permit_type
-          last_three_deprecated_template_versions
-          scheduled_template_versions
-          published_template_version
-        ],
         order: order,
         where: {
           discarded: discarded,
-          early_access: early_access,
+          visibility: visibility,
         },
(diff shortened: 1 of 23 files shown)


app/models/requirement_template.rb
```diff
@@ -55,6 +55,10 @@ def assignee
     raise NotImplementedError, "The assignee method is not implemented"
   end
 
+  def visibility
+    raise NotImplementedError, "The assignee method is not implemented"
```
LaurelOlson, line 59: fix error text

app/controllers/api/concerns/search/requirement_templates.rb
```diff
@@ -27,11 +27,19 @@ def perform_search
...
+    end
   end
 
   def search_params
-    params.permit(:query, :show_archived, :early_access, :page, :per_page, sort: %i[field direction])
+    params.permit(:query, :show_archived, :visibility, :page, :per_page, sort: %i[field direction])
+  end
+
+  def visibility
+    search_params[:visibility].split(",")
```
LaurelOlson, line 42: if visibility can be multiple items, this would break line 30 no?

LaurelOlson (approved):

PR #750, opened 2024-10-31
BPHH-2252 update role of user on any re-invite

bkmaibach:
## Description

When a user is re-invited, update the users role if it is allowed.
bonus: fix the invite result tag display centering.

### DIFF
--- a/app/frontend/components/shared/base/inputs/user-input.tsx
+++ b/app/frontend/components/shared/base/inputs/user-input.tsx
@@ -118,10 +118,10 @@ const IInviteResultTag = ({ bg, icon, text, ...rest }: IInviteResultTagProps) =>
       noOfLines={1}
       bg={bg}
       color={color}
-      {...rest}
       display="flex"
       alignItems="center"
       gap={2}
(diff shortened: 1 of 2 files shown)


app/frontend/components/shared/base/inputs/user-input.tsx
```diff
@@ -111,11 +111,20 @@ const IInviteResultTag = ({ bg, icon, text, ...rest }: IInviteResultTagProps) =>
...
-        <Text>{text}</Text>
-      </HStack>
+    <Tag
+      border="1px solid"
+      borderColor={color}
+      mb={2}
+      noOfLines={1}
+      bg={bg}
+      color={color}
+      {...rest}
```
LaurelOlson, line 121: `{...rest}` should always be the final prop. That way the parent component can override any of the base props. Right now, it is not possible for the parent to override anything below the ...rest

bkmaibach: woops good catch

sunto (approved):

PR #765, opened 2024-11-13
Sandbox Admin + overall functionality 

bkmaibach:
rework temporary sandbox persistence for super sandbox and other sandbox dependence

## Description

https://hous-bssb.atlassian.net/browse/BPHH-2202


app/frontend/components/domains/permit-application/index.tsx
```diff
@@ -38,19 +38,26 @@ export const PermitApplicationIndexScreen = observer(({}: IPermitApplicationInde
...
+    JSON.stringify(currentSandboxId),
+  ])
   useFlashQueryParam()
   const resetQueryParams = useResetQueryParams()
 
   return (
     <Flex as="main" direction="column" w="full" bg="greys.white" pb="24">
       <PermitApplicationStatusTabs />
       <BlueTitleBar title={t("permitApplication.indexTitle")} />
+      {currentSandboxId}
```
LaurelOlson, line 60: is this intentional or was it for testing?

app/frontend/stores/root-store.ts
```diff
@@ -88,11 +92,6 @@ export const RootStoreModel = types
       self.userChannelConsumer?.consumer?.disconnect()
     },
   }))
-  .actions((self) => ({
-    afterCreate() {
-      self.loadLocalPersistedData()
-    },
-  }))
```
LaurelOlson, line 95: removing this seems like it might have unintended consequences. e.g. for the `afterLoginPath` - looks like this needs to persist while the user is logged out?

app/frontend/components/domains/permit-application/index.tsx
bkmaibach: testing, good catch

app/frontend/stores/root-store.ts
bkmaibach: ah, I will leave this then.

LaurelOlson (approved):

PR #769, opened 2024-11-15
and sandbox id to api key, properly scope and authorize actions, fix specs

bkmaibach:
## Description

external API keys now specify a sandbox. If they do not specify a sandbox, they will behave as before obviating the need for any data migration.

- [ x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

https://hous-bssb.atlassian.net/browse/BPHH-2098


app/models/live_requirement_template.rb
```diff
@@ -1,19 +1,22 @@
...
     existing_record =
       LiveRequirementTemplate.find_by(
         permit_type_id: permit_type_id,
         activity_id: activity_id,
         first_nations: first_nations,
         discarded_at: nil
       )
-    if existing_record.present?
+
+    if existing_record.present? && existing_record.id != id
```
taw27, line 19: I think it might be better to change this to:

```
existing_record =
      LiveRequirementTemplate.where.not(id: id).find_by(
        permit_type_id: permit_type_id,
        activity_id: activity_id,
        first_nations: first_nations,
        discarded_at: nil
      )
    if existing_record.present?

```
I think the current implementation will work, but I’m not totally sure which record find_by would return first if there are two matches. So I think the suggestion is more explicit

taw27 (approved):

PR #795, opened 2024-11-27
Made changes so that super admin will be able to search for review managers using their email addresses

mrsasi7994:
Previously Super Admin Can't able to manager Review Managers using their email Addresses. Now made changes so that As a Super Admin, they can search Review Managers using their email.

### DIFF
--- a/app/models/jurisdiction.rb
+++ b/app/models/jurisdiction.rb
@@ -10,9 +10,9 @@ class Jurisdiction < ApplicationRecord
   ]
 
   include ActionView::Helpers::SanitizeHelper
-  searchkick searchable: %i[name reverse_qualified_name qualified_name],
-             word_start: %i[name reverse_qualified_name qualified_name],
-             text_start: %i[name reverse_qualified_name qualified_name]
+  searchkick searchable: %i[name reverse_qualified_name qualified_name, review_manager_emails],
+             word_start: %i[name reverse_qualified_name qualified_name, review_manager_emails],
+             text_start: %i[name reverse_qualified_name qualified_name, review_manager_emails]
 
   # Associations
   has_one :preference
@@ -74,6 +74,10 @@ class Jurisdiction < ApplicationRecord
     users&.kept&.review_manager
   end
 
+  def review_manager_emails
+    review_managers&.pluck(:email)
+  end
+
   def reviewers
     users&.kept&.reviewer
   end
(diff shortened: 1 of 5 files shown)


bkmaibach:
This code would let unauthenticated users to check if any email is associated with review managers on a jurisdiction. I'm not sure that is a wise thing to allow. Is there some specific reason we are pursuing this feature?

bkmaibach (changes requested):
Jurisdiction search is a public unauthenticated endpoint, and associated review managers is not public data. If this feature is absolutely necessary, please find a way to prevent the public from uncovering private information.

bkmaibach (changes requested):
Please:

- Use role checking methods where appropriate (.super_admin? etc)
- Handle the case of regional review managers
- apply syntax_tree formatting (see README)

app/models/jurisdiction.rb
```diff
@@ -10,10 +10,11 @@ class Jurisdiction < ApplicationRecord
...
 
   include ActionView::Helpers::SanitizeHelper
-  searchkick searchable: %i[name reverse_qualified_name qualified_name],
-             word_start: %i[name reverse_qualified_name qualified_name],
-             text_start: %i[name reverse_qualified_name qualified_name]
+  searchkick searchable: %i[name reverse_qualified_name qualified_name, review_manager_emails],
+             word_start: %i[name reverse_qualified_name qualified_name, review_manager_emails],
+             text_start: %i[name reverse_qualified_name qualified_name, review_manager_emails]
 
+  SEARCH_DATA_FIELDS = %i[qualified_name reverse_qualified_name regional_district_name name type updated_at review_managers_size reviewers_size permit_applications_size user_ids submission_inbox_set_up created_at review_manager_emails]
```
sunto, line 17: I would leave out updated_at and created_at

sunto (approved):

guri-dev (approved):

bkmaibach (approved):
Nicely done

app/models/jurisdiction.rb
```diff
@@ -130,7 +171,9 @@ def search_data
       permit_applications_size: permit_applications_size,
       user_ids: users.pluck(:id),
       submission_inbox_set_up: submission_inbox_set_up,
-      created_at: created_at
+      created_at: created_at,
+      review_manager_emails: review_manager_emails,
+      regional_review_manager_emails: regional_review_manager_emails
```
bkmaibach, line 176: This might be better if they were the same field, like manager_emails or something

bkmaibach (changes requested):

bkmaibach (approved):

PR #874, opened 2025-01-14
Change visual identifier of SHA to @SHA instead of -SHA

derekyau:
change `-SHA` to `@SHA`


smithskye (approved):
Approved - Only change was character replacement of - to @ symbol.

PR #909, opened 2025-03-26
some ammendments to inbox switch pages and controller

bkmaibach:
## Description

Some recommended changes for Guri's PR:

- update action does not need to perform input validation
- change the name of the component to be a bit more specific
- Use a grid instead of a table, do away with static pixel values for more responsiveness
- remove unecessary code from switchbutton, make it compatible wherever chakra switch is used
- cut down flexboxes into a simpler structure
- make more of the arrow icon row thing clickable


guri-dev (approved):
Everything looks good to me

PR #912, opened 2025-03-27
fix sandbox switching UX and warnings

bkmaibach:
## Description

- properly handle the error when a permit template does not load (as when sandbox is switched)
- improve sandbox modal warning messaging (switching sandbox will kill unsaved changes)


geekinc:
There's a fair bit of code/text copy in this PR, but it all looks good 😄

guri-dev (approved):
Changes look good to me

PR #913, opened 2025-03-28
omniauth experiment

bkmaibach:
## Description

Trying to figure out how to save id_token as a secondary token to the access token in jwt devise gem

trying to figure out how to redirect in api endpoint without causing CORS issues


app/controllers/api/omniauth_callbacks_controller.rb
```diff
@@ -4,12 +4,22 @@ class Api::OmniauthCallbacksController < Devise::OmniauthCallbacksController
...
     result =
       OmniauthUserResolver.new(
-        auth: request.env["omniauth.auth"],
+        auth: auth,
         invitation_token: origin_query["invitation_token"]
       ).call
+    binding.pry
+    # Store the ID token directly in the user object for use in jwt_payload
     @user = result.user
+    @user.keycloak_id_token = auth.extra.id_token if auth
```
bkmaibach, line 19: we dont want to do this most likely, we want to store the id_token (a jwt token) aongside the same way we store our main jwt token (as an http only cookie)

app/controllers/api/sessions_controller.rb
```diff
@@ -5,8 +5,41 @@ class Api::SessionsController < Devise::SessionsController
   skip_before_action :verify_signed_out_user
 
   def destroy
+    # Try to get the ID token from the JWT payload
+    id_token = nil
+    token = request.headers["Authorization"]&.split(" ")&.last
+
+    if token.present?
```
bkmaibach, line 12: try to obtain the token

app/controllers/api/sessions_controller.rb
```diff
@@ -5,8 +5,41 @@ class Api::SessionsController < Devise::SessionsController
...
+
+    # Clear the session
+    reset_session
+
+    # If there's an ID token, redirect to Keycloak logout
+    if id_token.present? && ENV["LOGINPROXY_LOGOUT_URL"].present?
+      redirect_url = ENV["VITE_POST_LOGOUT_REDIRECT_URL"] || root_url
+      logout_url =
+        "#{ENV["LOGINPROXY_LOGOUT_URL"]}?post_logout_redirect_uri=#{CGI.escape(redirect_url)}&id_token_hint=#{id_token}"
+      redirect_to logout_url, allow_other_host: true
```
bkmaibach, line 38: try to redirect? except this is an api endpoint
Laurel suggested doing a "head" here except it still produces CORS errors

PR #915, opened 2025-03-31
Hub 1893 bug when logging out the page is not redirected to bph login page url in response

bkmaibach:
## Description

Uses new VITE_LOGINPROXY_LOGOUT_URL (UPDATE IN ALL ENVIRONMENTS)
sets the new id token in the logout URL which is returned by the session destroy endpoint.


bkmaibach:
PRing a different branch

PR #916, opened 2025-03-31
Hub 1893 bug when logging out the page is not redirected to bph login page cookie

bkmaibach:
## Description

Obtain  id_token and set in logoutUrl in session destroy response

https://hous-bssb.atlassian.net/browse/HUB-1893


geekinc (approved):
This is a clean and elegant way to manage the token and the redirect URL.  The only question (and it's not with the code) is, do we need to update any environment variables on the TEST and PROD environments (they're referenced in sessions_controller.rb)?

PR #968, opened 2025-04-25
add template generation and import tasks for step code levels

bkmaibach:
## Description

Task for generating step code template (to be filled)
one filled, the data can be imported with the second task.

NOTE:
Each row must have both values filled for the import to work.
The row data will not overwrite any existing step code data.


guri-dev (approved):
lgtm

PR #969, opened 2025-04-25
add touches after template controller updates

bkmaibach:
## Description

The touches are necessary as simply removing associations do not count as touching
also fix a browser import error


guri-dev (approved):
lgtm

geekinc (approved):
This looks good - in a follow up, it might be good to figure out why the reload is required to keep published_template_version from being nil.

PR #981, opened 2025-05-06
HUB-1748 created api/audit_logs to feed kibana using new gem pack

guri-dev:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

## UI Accessibility Checklist

- [ ] Checked
      for accessibility errors?

## General Checklist

- [ ] 📝 Use descriptive commit messages


app/controllers/api/audit_logs_controller.rb
```diff
@@ -0,0 +1,30 @@
+class Api::AuditLogsController < Api::ApplicationController
+  skip_after_action :verify_policy_scoped, only: :index
+
+  def index
+    audit_logs = Audited::Audit.all.order(created_at: :desc)
```
bkmaibach, line 5: If the application generates a large number of audit trails, loading all of them at once could lead to performance problems (slow response times, high memory usage on the server). This is a common concern with audit log tables, as they can grow very large over time.

app/controllers/api/audit_logs_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    render json: audit_logs.as_json, status: :ok
```
bkmaibach, line 7: Consider creating a blueprint

app/controllers/api/audit_logs_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  rescue NameError => e
+    if e.message.include?("Audited::Audit")
+      render_error(
+        "audit_logs.load_error",
+        message_opts: {
+          error_message:
+            "Could not load Audited::Audit. Ensure the 'audited' gem is configured correctly."
```
bkmaibach, line 14: Is there a good reason for catching this particular scenario? Sounds like there's a story behind this.

app/controllers/api/audit_logs_controller.rb
```diff
@@ -0,0 +1,30 @@
+class Api::AuditLogsController < Api::ApplicationController
+  skip_after_action :verify_policy_scoped, only: :index
```
bkmaibach, line 2: is it safe to show all audit logs, regardless of user? Because that's what this line says to do, as far as I understand.

app/controllers/api/audit_logs_controller.rb
bkmaibach: Pagination likely required, if there is even a need to serve this index endpoint.

app/models/jurisdiction.rb
```diff
@@ -286,6 +287,10 @@ def template_version_customization(template_version, sandbox = nil)
 
   private
 
+  def inbox_enabled_changed?
```
bkmaibach, line 290: is there a good reason to have a non-standard name for this boolean in addition to the standard one, thus making the codebase one step more crufted up?

app/models/site_configuration.rb
```diff
@@ -67,6 +69,10 @@ def max_revision_reasons
     end
   end
 
+  def inbox_enabled_changed?
+    saved_change_to_inbox_enabled?
```
bkmaibach, line 73: is there a good reason to have a non-standard name for this boolean in addition to the standard one, thus making the codebase one step more crufted up?

PR #982, opened 2025-05-06
rework audit setup using elasticsearch, uuids, no endpoint

bkmaibach:
## Description

working theory:
Because kibana is already connected to our elasticsearch, we just need to index the data into elasticsearch and then set up the index in kibana.
Once this is deployed we will know if this is a sound approach.


guri-dev (approved):
lgtm

PR #1000, opened 2025-05-16
add synchronous json reports for step code

bkmaibach:
## Description

Convert step code report generation to JSON,
being sure to remove submitter info.

Because this report is so large, it may not work with more than a few step code records. In the future we may need to architect an async websokcet notification / generation job / new attachment model system, that generates reports asynchronously, then websockets down a download link after attaching and uploading the generated file.

We need to triple check that no submitter info ever gets attached to this report.


bkmaibach:
To be revisited later

PR #1020, opened 2025-05-28
Hub 3175 story create readyness reports

bkmaibach:
## Description

All current work on permit projects/step code changes


app/controllers/concerns/step_code_params_concern.rb
```diff
@@ -0,0 +1,43 @@
+module StepCodeParamsConcern
+  extend ActiveSupport::Concern
+
+  private
+
+  def step_code_params
```
sunto, line 6: it could be possible we maintain separate part 3 and part 9 params. 
There are nearly a number of params shared though....

db/migrate/20250526232142_add_project_details_to_step_codes.rb
sunto, line 1: I think we can leave this out if we focus on the design that keeps the projectItem as the wrapper

app/models/user.rb
```diff
@@ -66,6 +66,8 @@ class User < ApplicationRecord
   has_one :preference, dependent: :destroy
   accepts_nested_attributes_for :preference
 
+  has_many :step_codes, as: :parent, dependent: :destroy
```
sunto, line 69: we should move this into a permit project instead.  A step_code should always belong to some sort of projectItem/project which then has responsibilities about ownership and sharing, etc.

PR #1053, opened 2025-06-19
Hub 1046 only designated reviewer to send revisions

guri-dev:

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

## UI Accessibility Checklist

- [ ] Checked
      for accessibility errors?

## General Checklist

- [ ] 📝 Use descriptive commit messages


db/migrate/20250616210003_set_default_designated_reviewer_for_jurisdictions.rb
```diff
@@ -0,0 +1,11 @@
+class SetDefaultDesignatedReviewerForJurisdictions < ActiveRecord::Migration[
+  7.1
+]
+  def up
```
bkmaibach, line 4: In many database systems, especially modern versions of PostgreSQL which is common for Rails apps, adding a column with a default value is a metadata change. This means existing rows will return the default value when queried without needing to be physically updated.
Given that, the add_column in the first migration...

```
# db/migrate/20250616204159_add_designated_reviewer_to_jurisdictions.rb
add_column :jurisdictions,
           :allow_designated_reviewer,
           :boolean,
           default: false,
           null: false
```
           
...should be sufficient to make all existing jurisdictions have allow_designated_reviewer as false.
The second migration, which explicitly runs an UPDATE query, is likely redundant.
This second migration forces a write to every row in the jurisdictions table, which can be a slow and expensive operation on large tables, and is generally what the "metadata-only default" feature is designed to avoid.

db/migrate/20250616210003_set_default_designated_reviewer_for_jurisdictions.rb
```diff
@@ -0,0 +1,11 @@
...
+    execute "UPDATE jurisdictions SET allow_designated_reviewer = false;"
+  end
+
+  def down
+    # This data migration is not easily reversible.
```
bkmaibach, line 9: Removing this migration would also have the benefit of allowing reversibility

db/schema.rb
```diff
@@ -41,6 +41,28 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "audits", force: :cascade do |t|
```
bkmaibach, line 44: This is from a different branch, let's try to keep the schema file consistent with migrations yeah? It's easier to spot now that the diffs in schema.rb arent all garbled by the formatter.

bkmaibach (changes requested):

app/policies/permit_application_policy.rb
```diff
@@ -69,7 +69,15 @@ def generate_missing_pdfs?
...
 
   def finalize_revision_requests?
-    user.review_staff? && record.submitted?
+    return false unless user.review_staff? && record.submitted?
+
+    feature_enabled =
+      SiteConfiguration.allow_designated_reviewer? ||
+        record.jurisdiction.allow_designated_reviewer
+
+    return true unless feature_enabled
```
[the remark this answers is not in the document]
guri-dev: @bkmaibach  This has been reviewed and updated.

db/migrate/20250616210003_set_default_designated_reviewer_for_jurisdictions.rb
guri-dev: @bkmaibach  Updated the migration

db/schema.rb
guri-dev: @bkmaibach  Schema file has been updated with relevant commit

PR #1067, opened 2025-06-25
always make obtained customizations sandbox specific

bkmaibach:
## Description

- make obtaining customizations sandbox-specific in all scanarios (espcially pertaining to external API and mappings)
- fix store filter resetting edge case for permit application store
- refresh page after selecting sandbox


guri-dev (approved):
lgtm

PR #1100, opened 2025-07-09
Hub 3278 story create tools discovery page and basic layout

bkmaibach:
## Description

This adds pages for step code check and allows the user to start a step code outside of a permit application.
This is what was public demo'd Wednesday July 9

it also moves some fields and changes the way they are defaulted.

It adds logic for creating and fetching part 3 step codes.

also renamed checklist fields.

Also adds creator to project items so that authorization can be checked apart from any other parent.


guri-dev (approved):
lgtm

PR #1104, opened 2025-07-14
validate required steps

bkmaibach:
## Description

Some unvalidated records were causing havoc on TEST - this removes


guri-dev (approved):
lgtm

PR #1115, opened 2025-07-17
set select options in part 9 fetch

bkmaibach:
## Description

Fix bugs prevent load of step code configuration screen


guri-dev (approved):
lgtm

PR #1119, opened 2025-07-18
modify project item delegations and migrations to reenable pa creation

bkmaibach:
Notice that the migrations were changed which means they had to be reversed on develop. This stuff should not be anywhere else so that's fine.
This fixes the errors currently present on develop that require a project object when creating a permit application through the existing method. Delegations to project specific fields can now be overridden by the previously PA specific fields.


guri-dev (approved):
lgtm

PR #1132, opened 2025-07-24
HUB-3583 - Permit Type Required Steps

bkmaibach:
## Description

remove non-part-9 required steps and fix selector

bonus: fix required block name display for admin sidebar


guri-dev (approved):
lgtm

PR #1178, opened 2025-08-27
Hub 3734 bug webhook not working

bkmaibach:
## Description

Webhook was failing do to missing question mark.

Also fixed horrendous URL input component by changing the normalization to occur onblur instead of onchange.

Added specs for webhook service.


guri-dev (approved):
lgtm

PR #1184, opened 2025-09-05
remove extra navigate call 

bkmaibach:
so that form actually validates and saves on part 3 step code "save and go back"


guri-dev (approved):
lgtm

PR #1186, opened 2025-09-08
fix creation of classificaitons by migrating classification code to s…

bkmaibach:
by migrating classification code to string


guri-dev (approved):
lgtm

PR #1195, opened 2025-09-10
ensure pdf cleanup before and after generation, fix ref number update

bkmaibach:
## Description

fix pdf re-generation, ensure PDF cleanup, improve copy, fix reference number defaulting in step code


guri-dev (approved):
lgtm

PR #1202, opened 2025-09-17
use new endpoint for loading step code optons in step requirements screen

bkmaibach:
## Description

Fix bug where requirements screen is not loading due to using outdated options endpoint


guri-dev (approved):
lgtm

PR #1205, opened 2025-09-19
Hub 3557 UI / ux ux UI new requirement block options for entering the step codes part 3 workflow

bkmaibach:
## Description

add ability to select previous step codes


guri-dev (approved):
lgtm

PR #1223, opened 2025-10-03
use reflection as fallback for obtaining parent id column when not loaded

bkmaibach:
ProjectItem concern fails under cases where records are loaded using complex custom queries such as in the application metrics report. To get around this, the concern has_parent  method has to intelligently know wha to use:

- Reuse loaded associations and only lazy-load when the foreign key attribute exists; return `nil` otherwise. (returning nil in this case is fine because we know that any custom queries arent interested in this field, or else the foreign key would be present)
- Prevent `ActiveModel::MissingAttributeError` when exporting metrics (missing `permit_project_id`).

Bonus: Guard `ProjectItem` so it only mixes into ActiveRecord models.


guri-dev (approved):
lgtm

PR #1489, opened 2026-02-24
add matomo script

bkmaibach:
## Description

Add Matomo script to root component.
Several other configuration steps must be followed, see the Matomo React instructions.


mymobanty (approved):
lgtm

mymobanty:
> @mymobanty Just fyi - we havent approved this for development. I would like a presentation about it and how it would work and the other options you looked at.

so I was thinking we could POC it on dev to see how it works , else the presentation would just be on the functionality and perhaps using other data I can find in the public

PR #1612, opened 2026-04-10
HUB 4770 review manager can edit all new about page content sections

IanFonzie:
## Description

Edit new fields; layout touchups.

- [X] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📦 Chore (Release)
- [ ] ✅ Test

[HUB-4770](https://hous-bssb.atlassian.net/browse/HUB-4770)

## Steps to QA

Compare against the ticket

## UI Accessibility Checklist

- [X] Markup uses semantic HTML?
- [X] Additional context added through `aria-roles` and `aria-labels`?
- [X] Checked
      for accessibility errors?
- [X] Usable with a keyboard and navigable in a logical order?
- [X] Usable using a screen reader (e.g. Voiceover for Mac) with enough context for the user?

## General Checklist

- [X] ✅ Provide tests for your changes where relevant
- [X] 📝 Use descriptive commit messages
- [X] 📗 Update any related documentation and include any relevant screenshots
- [X] 📱 For UI-related tasks, ensure responsive design is implemented across multiple screen size
- [X] 🗂️ Move your relevant story/task to the _Ready for Code Review_ state

[HUB-4770]: https://hous-bssb.atlassian.net/browse/HUB-4770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/frontend/components/domains/jurisdictions/jurisdiction-screen.tsx
```diff
@@ -101,6 +100,30 @@ export const JurisdictionScreen = observer(() => {
     reset(getDefaultJurisdictionValues())
   }, [currentJurisdiction?.id])
 
+  const canManageAbout = can("jurisdiction:manage", { jurisdiction: currentJurisdiction })
```
bkmaibach: canManageAbout is computed before the early return guard. Line 103 calls can("jurisdiction:manage", { jurisdiction: currentJurisdiction }) before the if (!currentJurisdiction) return <LoadingScreen /> check on line 131. This works only because the RBAC rule was updated to accept undefined/null jurisdiction (returning false), but it's fragile — if someone reverts that RBAC change, this crashes. Worth flagging.

app/frontend/components/domains/jurisdictions/jurisdiction-screen.tsx
```diff
@@ -118,33 +141,26 @@ export const JurisdictionScreen = observer(() => {
...
-        {!currentJurisdiction.inboxEnabled && (
-          <Box my={8}>
-            <CustomMessageBox status={EFlashMessageStatus.warning} description={t("jurisdiction.notEnabled")} />
-          </Box>
-        )}
-      </Container>
-      {currentUser?.isReviewStaff || showAboutPage ? (
-        <>
-          <FormProvider {...formMethods}>
+    <FormProvider {...formMethods}>
```
bkmaibach: FormProvider now wraps the entire return including the <HeroBanner> and the non-about-page fallback (notUsingBPH path). This means the form context is available even in the "jurisdiction not using BPH" view, which is unnecessary and could cause confusion.

app/frontend/services/rbac-rules.ts
```diff
@@ -14,7 +14,7 @@ const reviewManagerRules = {
   static: [...sharedStaticRules, "user:invite", "application:download", "user:updateRole"],
   dynamic: {
     ...sharedDynamicRules,
-    "jurisdiction:manage": (currentUser: IUser, data: { jurisdiction: IJurisdiction }) =>
+    "jurisdiction:manage": (currentUser: IUser, data: { jurisdiction?: IJurisdiction | null }) =>
```
bkmaibach: The jurisdiction:manage dynamic rule signature changed from { jurisdiction: IJurisdiction } to { jurisdiction?: IJurisdiction | null }. This is a broadening change that affects all callers of this rule, not just the about page. The rule now gracefully returns false for null, but verify no existing callers relied on the previous guarantee that jurisdiction is always defined (i.e., would a false there hide a real bug elsewhere?).

app/frontend/components/domains/jurisdictions/jurisdiction-contact-info-banner.tsx
```diff
@@ -0,0 +1,349 @@
...
+  return digits ? `tel:${digits}` : `tel:${phone}`
+}
+
+interface IJurisdictionContactInfoBannerProps {
+  control: Control<TJurisdictionFieldValues>
+  canManage: boolean
+  qualifiedName: string
+}
+
+export function JurisdictionContactInfoBanner({
```
bkmaibach, line 150: This is the largest new component. The Controller rules prop uses a deeply nested ternary chain (key === "officeAddress" ? ... : key === "officeHours" ? ... : key === "officeTelephone" ? ... : ...) which is hard to read and maintain. Look at lines around the rules prop — four levels of ternary nesting for validation config. Consider if this should be a lookup map instead.

Also: the component has a hardcoded EMAIL_REGEX — check if it matches the backend's URI::MailTo::EMAIL_REGEXP. Mismatched frontend/backend validation is a common source of bugs.

See Teams message for suggested re-write

bkmaibach:
The extraction of useEditorWithPreview hook is clean, but the original EditorWithPreview component now delegates to the hook while the new JurisdictionEditorWithPreview also uses the same hook with a different UI. Check that the old EditorWithPreview (used elsewhere in the app) still behaves identically — the containerChrome: "default" path should match the old behavior exactly.

db/schema.rb
```diff
@@ -15,7 +15,10 @@
   enable_extension "pgcrypto"
```
bkmaibach, line 15: maybe do regen the schema.

bkmaibach (changes requested):
Not bad but a few things to carefully consider

IanFonzie:
> The extraction of useEditorWithPreview hook is clean, but the original EditorWithPreview component now delegates to the hook while the new JurisdictionEditorWithPreview also uses the same hook with a different UI. Check that the old EditorWithPreview (used elsewhere in the app) still behaves identically — the containerChrome: "default" path should match the old behavior exactly.

Went through and compared a few instances of them; they look the same to me

app/frontend/components/domains/jurisdictions/jurisdiction-screen.tsx
```diff
@@ -47,104 +43,119 @@ import { Can, can } from "../../shared/user/can"
...
+      contactInfoBannerHasContent(
+        currentJurisdiction?.officeAddress,
+        currentJurisdiction?.officeHours,
+        currentJurisdiction?.officeTelephone,
+        currentJurisdiction?.officeEmail
+      ) ||
+      (currentJurisdiction?.contacts?.length ?? 0) > 0
+    return (
+      <Flex as="main" direction="column" w="full" bg="greys.white">
+        <FormProvider {...formMethods}>
```
bkmaibach, line 154: The hero banner has its own FormProvider (line 136) and renders JurisdictionHeroLgWebsiteRow with an Edit/Done button whenever canManageAbout is true. The <form> with the Save button only renders under currentUser?.isReviewStaff || showAboutPage.
If a manager has permission (canManageAbout === true) but !isReviewStaff && !showAboutPage, they can enter a URL and click Done, but there is no Save button anywhere so the value is never persisted. Either gate the hero edit UI on the same condition, or move the Save bar out of the conditional branch.
Also: having two sibling FormProviders is unusual — they share formMethods so it works, but collapsing them into one by wrapping the outer <Flex> would be cleaner and avoid this class of bug.

app/models/jurisdiction.rb
```diff
@@ -89,6 +90,15 @@ class Jurisdiction < ApplicationRecord
...
   validates :locality_type, presence: true
+  validates :processing_time_html, length: { maximum: 140 }, allow_blank: true
+  validates :key_stages_html, length: { maximum: 140 }, allow_blank: true
+  validates :office_address, length: { maximum: 500 }, allow_blank: true
+  validates :office_email,
+            format: {
+              with: URI::MailTo::EMAIL_REGEXP
+            },
+            allow_blank: true
+  validates :office_telephone, phone: true, allow_blank: true
```
bkmaibach: app/models/jurisdiction.rb validates length for processing_time_html (140), key_stages_html (140), office_address (500). The frontend additionally caps office_hours at 500, office_telephone at 128, office_email at 255 and timeline_and_deliverables_html is unbounded on both sides. The server should enforce matching caps — a direct API call can otherwise persist arbitrarily long values.

bkmaibach (approved):

PR #1722, opened 2026-06-04
HUB-4635 add project readiness tools back to nav

IanFonzie:
Accidentally removed the wrong route; /project-readiness-tools/check-digital-seals is still there, it's just protected.


bkmaibach (approved):
Ship it

